PR #175, opened 2023-06-12
172 data table show additional info

amarkensteijn:


amarkensteijn:
Part of the unit test is disabled in 96c8d6f due to issues with the locale. Created to #184 deal with this.

PR #181, opened 2023-07-05
only use url if the mainpanel is not set to time series dialog

grijzea:


ekkelenkamp (changes requested):
"time series dialog" should be constant.

ekkelenkamp (approved):

PR #188, opened 2023-07-27
186 bounding box selection mapbox

martinapippi:


src/components/MapComponent.vue
```diff
@@ -1,5 +1,7 @@
 <template>
   <div class="map-container">
+    <link rel="stylesheet" href="https://api.mapbox.com/mapbox-gl-js/plugins/mapbox-gl-draw/v1.4.0/mapbox-gl-draw.css" type="text/css">
```
ceesvoesenek, line 3: Do we want to use a CDN for these stylesheets, or would we rather host these ourselves?

src/components/Regridder.vue
```diff
@@ -0,0 +1,232 @@
...
+    @Prop() firstValueTime!: string
+    @Prop() lastValueTime!: string
+
+    mapObject!: Map
+    isInitialized = false
+    draw!: MapboxDraw
+    downloadControl: DownloadControl | undefined
+    downloadDialog = false
+    errorDialog = false
+    dx = ''
```
ceesvoesenek, line 56: Provide sensible defaults for `dx` and `dy`?

src/components/Regridder.vue
```diff
@@ -0,0 +1,232 @@
+<template>
+    <div> 
+      <!-- Download Dialog -->
+      <v-dialog v-model="downloadDialog" max-width="500">
+        <v-card>
+          <v-card-title>Download netCDF Data</v-card-title>
+          <v-card-text>
+            <v-form>
+              <v-text-field v-model="dx" label="dx" required></v-text-field>
```
ceesvoesenek, line 9: Maybe label "Step size in x-direction" or something instead of "dx"?

src/components/Regridder.vue
```diff
@@ -0,0 +1,232 @@
...
+export default class Regridder extends Mixins(PiRequestsMixin) {
+    @Inject() getMap!: () => Map
+
+    @Prop() firstValueTime!: string
+    @Prop() lastValueTime!: string
+
+    mapObject!: Map
+    isInitialized = false
+    draw!: MapboxDraw
+    downloadControl: DownloadControl | undefined
```
ceesvoesenek, line 53: In my experience, reactivity on `undefined` variables will often confuse Vue and lead to errors. Using `null` solves these problems, as Vue then realises that the variable actually does exist but is just intentionally not set.

src/components/Regridder.vue
```diff
@@ -0,0 +1,232 @@
...
+            this.mapObject.removeControl(this.downloadControl)
+        }
+        this.downloadControl = new DownloadControl(bbox, this)
+        this.mapObject.addControl(this.downloadControl)
+    }
+
+    select() {
+        const data = this.draw.getAll()
+        // get bbox from data
+        const geometry: any = data.features[0].geometry // change any  to proper type
```
ceesvoesenek, line 90: Yes, please 😄

src/components/Regridder.vue
```diff
@@ -0,0 +1,232 @@
...
+        if (geometry.coordinates !== null) {
+            const coordinates = geometry.coordinates[0]
+            const x = coordinates.map((c: any) => c[0])
```
ceesvoesenek, line 93: Please also replace these `any` when you add the type above

src/components/Regridder.vue
```diff
@@ -0,0 +1,232 @@
...
+        defaultMode: 'draw_rectangle',
+        })
+
+        this.mapObject.addControl(this.draw)
+
+        this.draw.changeMode("draw_rectangle")
+
+        this.mapObject.on('draw.create', this.select)
+
+        const trashElement = document.querySelector(".mapbox-gl-draw_trash")
```
ceesvoesenek, line 123: Select on `id` rather than `class`? `id` should always be unique, while multiple elements may share the same `class`

src/components/Regridder.vue
```diff
@@ -0,0 +1,232 @@
...
+    this.errorDialog = false
+    // reset error message
+    this.errorMessage = 'Select an area on the map before downloading the data.'
+    }
+
+    _downloadNetCDF(apiUrl: string) {
+        let fileName = 'openarchive-netcdf.nc'
+        // Use the Fetch API to make the request, change later to fews-pi-requests
+        fetch(apiUrl)
+        .then((response) => {
```
ceesvoesenek, line 194: Use `async` function and `await`? It's a bit more easy to read and more consistent with the rest of `fews-web-oc`.

src/components/Regridder.vue
```diff
@@ -0,0 +1,232 @@
...
+
+    closeErrorDialog() {
+    this.errorDialog = false
+    // reset error message
+    this.errorMessage = 'Select an area on the map before downloading the data.'
+    }
+
+    _downloadNetCDF(apiUrl: string) {
+        let fileName = 'openarchive-netcdf.nc'
+        // Use the Fetch API to make the request, change later to fews-pi-requests
```
ceesvoesenek, line 192: Is there something stopping us from changing it to `fews-pi-requests` now?

src/lib/DownloadControl.ts
```diff
@@ -0,0 +1,60 @@
+import {Map} from 'mapbox-gl'
+import Vue from 'vue'
+
+export class DownloadControl {
+  bbox: any | undefined
+  private _map: Map | undefined
```
ceesvoesenek, line 6: The underscore prefix is not necessary

src/lib/DownloadControl.ts
```diff
@@ -0,0 +1,60 @@
...
+  constructor(bbox: any, vueComponent: Vue) {
+    this.bbox = bbox
+    this._vueComponent = vueComponent
+  }
+
+  onAdd(map: Map) {
+    this._map = map
+
+    this._btn = document.createElement('button')
+    this._btn.className = 'mapboxgl-ctrl-download'
```
ceesvoesenek, line 20: Also add an `id` here, as mentioned above.

src/lib/DownloadControl.ts
```diff
@@ -0,0 +1,60 @@
...
+    this._btn.type = 'button'
+
+    const icon = document.createElement('i')
```
ceesvoesenek, line 23: Why is the icon in a `<i>` tag?

src/lib/DownloadControl.ts
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  onRemove() {
+    if (this._container && this._container.parentNode) {
+      this._container.parentNode.removeChild(this._container)
+    }
+    this._map = undefined
+  }
+
+  private _showPopup() {
```
ceesvoesenek, line 49: Also here, underscores are not necessary and make the code a bit more difficult to read and inconsistent with the rest of `fews-web-oc`, in my opinion

src/components/Regridder.vue
```diff
@@ -0,0 +1,232 @@
...
+        const startTime = this.firstValueTime
+        const endTime = this.lastValueTime
+        // todo: change this with fews-pi-requests
+        const apiUrl = `${this.baseUrl}rest/fewspiservice/v1/processdata?workflowId=${workflowId}&xMin=${xMin}&yMin=${yMin}&xMax=${xMax}&yMax=${yMax}&xCellSize=${xCellSize}&yCellSize=${yCellSize}&startTime=${startTime}&endTime=${endTime}`
+        this._downloadNetCDF(apiUrl)
+        this.downloadDialog = false
+    } else {
+        // Handle invalid input
+        this.closeDownloadDialog()
+        this.openErrorDialog('Invalid input. Please enter valid values for dx and dy.')
```
ceesvoesenek, line 171: It may be a bit annoying that the download dialog is closed upon error, because users have to click the download button again if they made an error.

src/lib/DownloadControl.ts
```diff
@@ -0,0 +1,60 @@
...
+
+    const icon = document.createElement('i')
+    icon.className = 'fas fa-download'
+
+    this._btn.appendChild(icon)
+    this._btn.onclick = () => {
+      if (this.bbox !== undefined) {
+        this._showPopup()
+      } else {
+        this._showErrorPopup()
```
ceesvoesenek, line 31: Maybe just disable the button when the bounding box is not yet defined, instead of showing an error?

ceesvoesenek (commented):
Cool feature 😄 Also nice integration with MapBox for the control.

Most important comments:
- Should we be using stylesheets from a CDN?
- Hardcoded parameters for the WMS request should be fixed
- Can we use `fews-pi-requests`?

Minor style issues:
- We don't prefix `private` methods and properties with an `_` in `fews-web-oc`
- Indentation is a bit weird in many places, maybe run an autoformatter (with indent size 2)

src/lib/DownloadControl.ts
martinapippi: I thought that it was the "normal" approach https://www.geeksforgeeks.org/how-to-add-icons-in-the-button-in-html/

src/components/MapComponent.vue
martinapippi: used mdi instead

src/components/Regridder.vue
```diff
@@ -0,0 +1,232 @@
...
+        this.isInitialized = true
+        this.addToMap()
+    }
+    }
+    downloadClicked() {
+    const dx = parseFloat(this.dx)
+    const dy = parseFloat(this.dy)
+
+    if (!isNaN(dx) && !isNaN(dy) && dx > 0 && dy > 0 && this.bbox !== undefined)  {
+        const workflowId = 'Transformation_ASA_Grid_Wind'
```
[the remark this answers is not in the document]
martinapippi: it's hardcoded until we get a `workflowId ` from GetCapabilities (probably it will be the layer name)

src/components/Regridder.vue
martinapippi: unfortunately that doesn't seem to be possible since the trashElement is defined by mapbox

PR #191, opened 2023-08-08
189 front end implementation default zoom extent

martinapippi:


src/views/SpatialDisplay.vue
```diff
@@ -47,10 +48,15 @@ import { ColourMap } from '@deltares/fews-web-oc-charts'
...
 import { Layer } from '@deltares/fews-wms-requests'
 import {LayerGroup} from "@deltares/fews-wms-requests/src/response/getCapabilitiesResponse";
-
-interface MapboxLayerOptions {
-  name: string;
-  time: Date;
+import { toWgs84 } from '@turf/projection';
+import { point } from '@turf/helpers';
+
+interface BoundingBox {
```
amarkensteijn, line 54: Can we use the `BoundingBox` interface from [@deltares/fews-wms-request](https://github.com/Deltares/fews-wms-requests/blob/e2a8c0134ad47963a347cde660852ba438e99abe/src/response/getCapabilitiesResponse.ts#L70)?

src/views/SpatialDisplay.vue
```diff
@@ -127,9 +134,31 @@ export default class SpatialDisplay extends Mixins(WMSMixin) {
...
+
+    const minx = parseFloat(boundingBox.minx)
+    const miny = parseFloat(boundingBox.miny)
+    const maxx = parseFloat(boundingBox.maxx)
+    const maxy = parseFloat(boundingBox.maxy)
+
+    const p1 = toWgs84(point([minx, miny], { crs: crs }))
+    const p2 = toWgs84(point([maxx, maxy], { crs: crs }))
+
+    return [
```
amarkensteijn, line 154: Have it return a [Mapbox LngLatBoundsLike]( https://docs.mapbox.com/mapbox-gl-js/api/geography/#lnglatboundslike), as that is the type that `fitBounds()` expects.

src/components/AnimatedMapboxLayer.vue
```diff
@@ -22,9 +21,21 @@ function getCoordsFromBounds(bounds: LngLatBounds) {
   ]
 }
 
-interface MapboxLayerOptions {
+function isBoundsWithinBounds(innerBounds: LngLatBounds, outerBounds: LngLatBounds) {
```
amarkensteijn, line 24: Maybe use [Turf's booleanWithin](http://turfjs.org/docs/#booleanWithin) to check if one bounds is within another bounds?

src/views/SpatialDisplay.vue
amarkensteijn, line 1: Similar changes need to be applied to de DataView.vue

amarkensteijn (changes requested):

src/components/AnimatedMapboxLayer.vue
martinapippi: Type definitions of `booleanWithin` are missing furthermore the `innerBounds` and `outerBounds` should be transformed into turf points and the function would have a similar amount of lines. This would have been the function, but it does not have an added value for me.
```
function isBoundsWithinBounds(innerBounds: LngLatBounds, outerBounds: LngLatBounds) {
  // Convert the bounds to Turf.js points
  const innerNW = point([innerBounds.getNorthWest().lng, innerBounds.getNorthWest().lat])
  const innerSE = point([innerBounds.getSouthEast().lng, innerBounds.getSouthEast().lat])
  const outerNW = point([outerBounds.getNorthWest().lng, outerBounds.getNorthWest().lat])
  const outerSE = point([outerBounds.getSouthEast().lng, outerBounds.getSouthEast().lat])

  const isInnerWithinOuter = booleanWithin(innerNW, outerNW) && booleanWithin(innerSE, outerSE)

  return isInnerWithinOuter
}
```

PR #193, opened 2023-08-09
Dwo 264 front end implementation web operation client configuration 

ekkelenkamp:


src/views/LoginView.vue
```diff
@@ -1,40 +1,65 @@
...
+  components: {
+    DeltaresLogin
+  }
+})
+export default class LoginView extends Vue {
+  webServiceUrl: string = ''
+  title: string = 'Delft-FEWS Web Operator Client'
+
+  async mounted(): Promise<void> {
+    const baseUrl = this.$config.get('VUE_APP_FEWS_WEBSERVICES_URL')
```
martinapippi, line 29: this method is a bit long, it can be split in multiple methods, something like this

```
async initializeComponent(): Promise<void> {
  await this.setWebOcConfiguration()
  await this.setTitle()
  await this.setFavicon()
  await this.setBackgroundImage()
  document.title = this.title
```

martinapippi:
image
The logo appears to have a non transparent background, maybe try with a different image?

PR #199, opened 2023-08-14
configure nginx.conf

ekkelenkamp:


wkramer:
The CSP header style-src inline is needed for the fews-web-oc-charts and not due to Vue.js.
In fews-web-oc-charts we use inline style in parsed html for the tooltip. We should change this with a trusted Javascript that creates the DOM elements.

PR #214, opened 2023-09-12
METOC:show time series when double clicking on wms layer

martinapippi:
Still to do:

- [ ] In `MetocDataView` execute `onCoordinatesChange` only when the coordinates change, add another watch for when the elevation changes and call `updateTimeSeriesWithElevation`  . This should improve the performance of the slider. It should look something like this, (but this doesn't work)
```
  @Watch('currentElevation')
  onCurrentElevationChange(): void {
    if (!this.currentElevation) return
    for (const resourceId in this.timeSeriesStore) {
      const series = this.timeSeriesStore[resourceId];
      this.updateTimeSeriesWithElevation(series, this.currentElevation)
      Vue.set(this.timeSeriesStore, resourceId, series)
    }
  }
```
- [x] there seem to be some problems with coordinates route, some errors are printed and if the coordinates values are inserted via the searchbar the plots are not shown.
- [x] whenever the user switches between layers, the window with the coordinates plot is not closed.
- [x] whenever the user clicks on coordinates on the map a point should appear. 
- [ ] set coordinates as title
- [ ] show vertical profile
- [ ] update chart legend

### DIFF
--- a/src/components/AnimatedMapboxLayer.vue
+++ b/src/components/AnimatedMapboxLayer.vue
@@ -182,11 +182,11 @@ export default class AnimatedMapboxLayer extends  Vue {
   }
 
   setDoubleClickLayer() {
-    if (this.mapObject) {
+    if (this.mapObject && this.mapObject.doubleClickZoom.isEnabled()) {
       // deactivate double click zoom
       this.mapObject.doubleClickZoom.disable()
       this.mapObject.on("dblclick", (e) => {
-        this.$emit("doubleClick", e)       
+        this.$emit("doubleclick", e)
       })
     }
   }
--- a/src/components/AnimatedMapboxLayer.vue
+++ b/src/components/AnimatedMapboxLayer.vue
@@ -181,10 +181,12 @@ export default class AnimatedMapboxLayer extends  Vue {
     }
   }
 
-  setDoubleClickLayer() {
-    if (this.mapObject && this.mapObject.doubleClickZoom.isEnabled()) {
-      // deactivate double click zoom
-      this.mapObject.doubleClickZoom.disable()
+  enableDoubleClickLayer() {
+    if (this.mapObject) {
+      if(this.mapObject.doubleClickZoom.isEnabled()){
+        // deactivate double click zoom
+        this.mapObject.doubleClickZoom.disable()
+      }
       this.mapObject.on("dblclick", (e) => {
         this.$emit("doubleclick", e)
       })
@@ -213,7 +215,7 @@ export default class AnimatedMapboxLayer extends  Vue {
     this.newLayerId = getFrameId(this.layer.name, this.counter)
     const source = this.mapObject.getSource(this.newLayerId)
     const baseUrl = this.$config.get("VUE_APP_FEWS_WEBSERVICES_URL")
-    this.setDoubleClickLayer()
+    this.enableDoubleClickLayer()
     if (this.currentLayer !== originalLayerName) {
       // set default zoom only if layer is changed
       this.setDefaultZoom()
--- a/src/components/AnimatedMapboxLayer.vue
+++ b/src/components/AnimatedMapboxLayer.vue
@@ -81,7 +81,7 @@ function getMercatorBboxFromBounds(bounds: LngLatBounds): number[] {
 }
 
 @Component
-export default class AnimatedMapboxLayer extends Vue {
+export default class AnimatedMapboxLayer extends  Vue {
   @Prop({
     default: () => {
       return null
@@ -181,6 +181,16 @@ export default class AnimatedMapboxLayer extends Vue {
     }
   }
 
+  setDoubleClickLayer() {
+    if (this.mapObject) {
+      // deactivate double click zoom
+      this.mapObject.doubleClickZoom.disable()
+      this.mapObject.on("dblclick", (e) => {
+        this.$emit("doubleClick", e)       
+      })
+    }
+  }
+
   @Watch("layer")
   onLayerChange(): void {
     if (!this.isInitialized) return
@@ -203,6 +213,7 @@ export default class AnimatedMapboxLayer extends Vue {
     this.newLayerId = getFrameId(this.layer.name, this.counter)
     const source = this.mapObject.getSource(this.newLayerId)
     const baseUrl = this.$config.get("VUE_APP_FEWS_WEBSERVICES_URL")
+    this.setDoubleClickLayer()
     if (this.currentLayer !== originalLayerName) {
       // set default zoom only if layer is changed
       this.setDefaultZoom()
--- a/src/components/AnimatedMapboxLayer.vue
+++ b/src/components/AnimatedMapboxLayer.vue
@@ -182,11 +182,12 @@ export default class AnimatedMapboxLayer extends  Vue {
   }
 
   enableDoubleClickLayer() {
-    if (this.mapObject) {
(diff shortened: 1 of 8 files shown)


src/components/AnimatedMapboxLayer.vue
```diff
@@ -181,6 +181,16 @@ export default class AnimatedMapboxLayer extends Vue {
     }
   }
 
+  setDoubleClickLayer() {
+    if (this.mapObject && this.mapObject.doubleClickZoom.isEnabled()) {
```
ceesvoesenek, line 185: Why is this a condition for enabling listening for double clicks?

src/components/AnimatedMapboxLayer.vue
```diff
@@ -181,6 +181,16 @@ export default class AnimatedMapboxLayer extends Vue {
     }
   }
 
+  setDoubleClickLayer() {
```
ceesvoesenek, line 184: Maybe rename to something like `enableDoubleClickListener`?

src/lib/TimeSeries/timeSeries.ts
```diff
@@ -36,6 +36,9 @@ export class Series {
   start?: Date;
   end?: Date;
   lastUpdated?: Date;
+  domains?: any[];
```
ceesvoesenek, line 39: Can we add an explicit type to this?

src/lib/Topology/timeseriesdisplay.ts
```diff
@@ -34,19 +21,19 @@ async function getFilterActions(provider: PiWebserviceProvider, filterId: string
...
-      provider, filterId, locationId
-    )
-    displays = displays.concat(displaysCur)
-    requests = requests.concat(requestsCur)
+  for (const filter of filters) {
+      const [displaysCur, requestsCur] = await fetchTimeSeriesDisplaysAndRequestsForSingleFilter(
+        provider, filter
+      )
+      displays = displays.concat(displaysCur)
+      requests = requests.concat(requestsCur)
```
ceesvoesenek, line 36: Nitpick: we indent with 2 spaces.. auto-format maybe?

src/lib/Topology/timeseriesdisplay.ts
```diff
@@ -1,29 +1,16 @@
...
-  const baseUrl = provider.locationsUrl({}).href.replace('/locations', '')
-  const actionsUrl = new URL(`${baseUrl}/filters/actions`)
-  actionsUrl.searchParams.append('filterId', filterId)
-  actionsUrl.searchParams.append('locationIds', locationId)
-
-  const response = await fetch(actionsUrl)
-  const data: ActionsResponse = await response.json()
+type FilterArray = (timeSeriesGridActionsFilter | filterActionsFilter)[];
+// guard fuctions, needed because is not possible to use instanceof/ typeof on interfaces
+function isFilterActionsFilter(filter: timeSeriesGridActionsFilter | filterActionsFilter): filter is filterActionsFilter {
```
ceesvoesenek, line 8: Cool! Didn't know about this

src/lib/TimeSeries/timeSeries.ts
ceesvoesenek, line 39: Why do we need these new properties now, and did we not need them before?

src/mixins/TimeSeriesMixin.ts
```diff
@@ -68,32 +67,67 @@ export default class TimeSeriesMixin extends Mixins(PiRequestsMixin) {
...
+                return {
+                  x: new Date( parsePiDateTime(event, timeZone)),
+                  y: event.value === missingValue ? null : +event.value
+                }
+              })
+            } else if (elevation && timeSeries.domains !== undefined && timeSeries) {
+              series.domains = timeSeries.domains
+              this.updateTimeSeriesWithElevation(series, elevation)
+            } else {
+              throw new Error('No data found')
```
ceesvoesenek, line 98: Make the error a bit more descriptive? Maybe mention that we failed to fetch a time series because the response is empty

src/lib/TimeSeries/timeSeries.ts
ceesvoesenek: Seems to be `[string][]` (i.e. array of 1-tuples with strings)

src/mixins/TimeSeriesMixin.ts
```diff
@@ -68,32 +67,67 @@ export default class TimeSeriesMixin extends Mixins(PiRequestsMixin) {
...
+    }
+    const domainAxisValues = timeSeries.domains[0].domainAxisValues
+    if (domainAxisValues !== undefined) {
+      const domain= domainAxisValues[0]
+      // convert domain.values to an array of numbers
+      domain.values = domain.values.map((value: [string]) => {
+        return +value[0]
+      })
+      // find the index of the closest number to the elevation
+      const elevationIndex = domain.values.reduce((prev: number, curr: number, index: number) => {
```
ceesvoesenek, line 119: This seems a bit more complicated than just computing the distance with a simple `map` and then taking `Math.min(...distance)` but that might be a matter of taste.

src/mixins/TimeSeriesMixin.ts
```diff
@@ -68,32 +67,67 @@ export default class TimeSeriesMixin extends Mixins(PiRequestsMixin) {
...
+        return (Math.abs(curr - elevation) < Math.abs(domain.values[prev] - elevation) ? index : prev)
+      }, 0)
+
+      const events = timeSeries.domains.slice(1)
+      timeSeries.data = events.map((event: any) => {
```
ceesvoesenek, line 124: Add type? It should even be possible to just infer

src/mixins/TimeSeriesMixin.ts
```diff
@@ -68,32 +67,67 @@ export default class TimeSeriesMixin extends Mixins(PiRequestsMixin) {
...
+        const e = event.events[0]
```
ceesvoesenek, line 125: Why do we do this? Do we always need the first element?

This could do with a comment.

src/components/AnimatedMapboxLayer.vue
vanmeeuwen: initially because of disabling the double click I guess. This is moved to another check inside this if

hvangeffen: The event used to be called multiple times when switching to a different data layer. This check makes it only be added the first time.

src/mixins/TimeSeriesMixin.ts
```diff
@@ -68,32 +67,67 @@ export default class TimeSeriesMixin extends Mixins(PiRequestsMixin) {
...
+              throw new Error('No data found')
+            }
           }
           Vue.set(this.timeSeriesStore, resourceId, series)
         }
       })
     }
   }
+
+  updateTimeSeriesWithElevation (timeSeries: Series, elevation: number): void {
```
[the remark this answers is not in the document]
vanmeeuwen: Actually it does update the timeseries

src/lib/TimeSeries/timeSeries.ts
hvangeffen: is `Domain[]`, added

src/mixins/TimeSeriesMixin.ts
hvangeffen: `slice(1)` skips the first element

PR #215, opened 2023-09-12
Units in legends

vanmeeuwen:


martinapippi:
LGMT! Is it fine that we don't have anymore the info hover on the title?

src/views/MetocDataView.vue
```diff
@@ -38,11 +38,6 @@
             :items="dataSources"
             @input="onSelectDataSource"
           />
-          <WMSInfoPanel
```
martinapippi, line 41: `WMSInfoPanel` is not in any other part of the code so we can delete it

vanmeeuwen: Since this is a branch from the WebOC where it will be used, I was not sure if I should remove it indeed

vanmeeuwen:
> LGMT! Is it fine that we don't have anymore the info hover on the title?

Tom agreed to remove it, explicitly asked on the additional info that won't be visible.

PR #232, opened 2023-09-15
229 regridder color of bouding box should be always in contrast with the changing background

hvangeffen:


martinapippi (changes requested):
It is not very visible in this case: 
image
with these changes
```
    const customStyles = [
      {
        id: 'gl-draw-polygon-fill-inactive',
        type: 'fill',
        filter: ['all', ['==', 'active', 'false'],
          ['==', '$type', 'Polygon'],
          ['!=', 'mode', 'static']
        ],
        paint: {
          'fill-color': '#c2bebe',
          'fill-outline-color': '#626262',
          'fill-opacity': 0.5
        }
      },
      {
        id: 'gl-draw-polygon-fill-active',
        type: 'fill',
        filter: ['all', ['==', 'active', 'true'],
          ['==', '$type', 'Polygon']
        ],
        paint: {
          'fill-color': '#c2bebe',
          'fill-outline-color': '#626262',
          'fill-opacity': 0.5
        }
```


it looks like this:
image

vanmeeuwen:
Using these colors it still is not always visible:
image

hvangeffen:
> Using these colors it still is not always visible: image

image
For me it looks quite clear

PR #233, opened 2023-09-15
FIX: 218 allow slider values inbetween marks

hvangeffen:
Fix the min, max and interval errors and simplify the logic of the slider code.

### DIFF
--- a/src/components/ElevationSlider.vue
+++ b/src/components/ElevationSlider.vue
@@ -35,16 +35,16 @@ function roundToNearest100 (x: number) {
   }
 })
 export default class ElevationSlider extends Vue {
-  @Prop({ default: 1 }) value!: number
+  @Prop({ default: -1 }) value!: number
   @Prop({ default: 0 }) minValue!: number
-  @Prop({ default: 10 }) maxValue!: number
+  @Prop({ default: -10 }) maxValue!: number
 
   currentValue: number = 0
   readonly numberOfMarks: number = 8
   marks: number[] = []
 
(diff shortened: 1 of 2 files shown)


src/components/ElevationSlider.vue
```diff
@@ -35,68 +35,61 @@ function roundToNearest100 (x: number) {
...
-  @Prop({ default: 0 }) maxValue!: number
+  @Prop({ default: 10 }) maxValue!: number
 
   currentValue: number = 0
   readonly numberOfMarks: number = 8
   marks: number[] = []
 
-  mounted() {
+  beforeMount() {
     this.onValueChange()
```
vanmeeuwen, line 47: The onValueChange on the beforeMount may give issues if we later will add logic to the onValueChange routine. Therefore setting a default value is better to do inline I think

src/components/ElevationSlider.vue
```diff
@@ -35,68 +35,61 @@ function roundToNearest100 (x: number) {
   }
 })
 export default class ElevationSlider extends Vue {
-  @Prop({ default: 0 }) value!: number
+  @Prop({ default: 1 }) value!: number
   @Prop({ default: 0 }) minValue!: number
-  @Prop({ default: 0 }) maxValue!: number
+  @Prop({ default: 10 }) maxValue!: number
```
vanmeeuwen, line 41: Shall we have default -10 to 0? with default value -1?
We know that the elevation is always negative.

PR #282, opened 2023-09-27
240 update the chart legend and title when a coord is selected

### LINKED ISSUE #240: Update the chart legend and title when a coord is selected

vanmeeuwen:
closes #240

### DIFF
--- a/src/mixins/TimeSeriesMixin.ts
+++ b/src/mixins/TimeSeriesMixin.ts
@@ -90,6 +90,8 @@ export default class TimeSeriesMixin extends Mixins(PiRequestsMixin) {
                   y: event.value === missingValue ? null : +event.value
                 }
               })
+            } else {
+              throw new Error('Timeseries data should include events')
             }
           }
           Vue.set(this.timeSeriesStore, resourceId, series)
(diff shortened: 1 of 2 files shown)


hvangeffen:
> I get that this does not use a vertical profile anymore, but none of the chart's legends or titles have changed with this new version.
> # Base
> 
> image
> # After this PR's changes
> 
> image

Something weird happened here
Today after full refresh it does indeed show better title's

image

src/mixins/TimeSeriesMixin.ts
```diff
@@ -84,56 +83,18 @@ export default class TimeSeriesMixin extends Mixins(PiRequestsMixin) {
             series.header.source = header.moduleInstanceId
             series.start = new Date(parsePiDateTime(header.startDate, timeZone) )
             series.end = new Date(parsePiDateTime(header.endDate, timeZone) )
-            if (!elevation && timeSeries.events !== undefined) {
+            if (timeSeries.events !== undefined) {
```
hvangeffen, line 86: Should the error also be removed here?

hvangeffen (approved):

src/mixins/TimeSeriesMixin.ts
vanmeeuwen: fair point

PR #284, opened 2023-09-27
248 vue3 timeseriesdisplay

wkramer:


amarkensteijn:
Know issue: No scroll option on Charts and Table. If too many charts, or too many entries in the table, part is not visible. Created #294 for this.

PR #285, opened 2023-09-27
283 bounding box layout

### LINKED ISSUE #283: Bounding box layout
- [x] Lon and Lat are interchanged in the bounding box
- [x] By default show only 1 decimal

vanmeeuwen:
closes #283

### DIFF
--- a/src/components/Regridder.vue
+++ b/src/components/Regridder.vue
@@ -35,8 +35,8 @@
                   <v-col cols="3">
                     <v-text-field v-model.number="bbox[0]"
                                   type="number"
-                                  label="Min. Latitude"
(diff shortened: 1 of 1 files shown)


hvangeffen:
image

When changing individual values it is now quite difficult to understand how the bbox changes.
For instance, the top right input now raises the bottom edge of the bbox.
Would this be a problem for the user?

vanmeeuwen:
> image
> 
> When changing individual values it is now quite difficult to understand how the bbox changes. For instance, the top right input now raises the bottom edge of the bbox. Would this be a problem for the user?

I don't think this is a problem. The user is used to this coordinate system. If it is unclear, we can change this based on the feedback.

PR #288, opened 2023-09-27
281 have csv filename include coordinates in lat and long

### LINKED ISSUE #281: Have CSV filename include coordinates in lat and long

hvangeffen:
closes #281

### DIFF
--- a/src/router/index.ts
+++ b/src/router/index.ts
@@ -87,7 +87,7 @@ export const routesViews: Array<RouteConfig> = [
         meta: { authorize: [], sidebar: true }
       },
       {
-        path: 'coordinates/:xCoord/:yCoord',
+        path: 'coordinates/:longitude/:latitude',
         name: 'MetocDataViewerWithCoordinates',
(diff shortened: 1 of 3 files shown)


src/router/index.ts
```diff
@@ -87,7 +87,7 @@ export const routesViews: Array<RouteConfig> = [
         meta: { authorize: [], sidebar: true }
       },
       {
-        path: 'coordinates/:xCoord/:yCoord',
+        path: 'coordinates/:longitude/:latitude',
```
vanmeeuwen, line 90: A short internet search puts latitude before longitude

vanmeeuwen (approved):

src/router/index.ts
hvangeffen: The most important is what the user of this would expect / want to see.
Currently this follows the order used by the mapbox internals but we could always work around this.

vanmeeuwen: We can keep it like this for now, awaiting the clients feedback

PR #302, opened 2023-09-29
Add key with sequence to get correct timeseries for a grid action

vanmeeuwen:


src/lib/ChartConfig/timeSeriesDisplayToChartConfig.ts
```diff
@@ -1,13 +1,14 @@
...
 import { ChartSeries } from "@/components/TimeSeriesComponent/lib/ChartSeries.js"
-import { TimeSeriesDisplaySubplot, TimeSeriesDisplaySubplotItem } from "@deltares/fews-pi-requests"
 import { cssStyleFromFewsLine, cssStyleFromFewsMarker, chartMarkerFromFews } from "./Styles"
+import { TimeSeriesDisplaySubplot, TimeSeriesDisplaySubplotItem } from "../TimeSeries/types"
 
 export function timeSeriesDisplayToChartConfig(subplot: TimeSeriesDisplaySubplot, title: string): ChartConfig {
   const config: ChartConfig = {
     title: title,
-    xAxis: [],
+    sequence: subplot.items[0].sequence !== undefined ? subplot.items[0].sequence : undefined,
```
amarkensteijn, line 10: Assumes there is never more than one item per subplot (which might be true for this use case)

src/lib/TimeSeries/types.ts
```diff
@@ -1,3 +1,6 @@
+import { TimeSeriesDisplaySubplotItem as SubPlotItem,
+  TimeSeriesDisplaySubplotItemAxis } from "@deltares/fews-pi-requests";
```
amarkensteijn, line 2: Add type to import:
```suggestion
import type { TimeSeriesDisplaySubplotItem as SubPlotItem,
  TimeSeriesDisplaySubplotItemAxis } from "@deltares/fews-pi-requests";

```

src/lib/TimeSeries/types.ts
```diff
@@ -24,3 +27,13 @@ export enum SeriesOperationType {
...
+
+export interface TimeSeriesDisplaySubplotItem extends SubPlotItem{
+  sequence?: number;
+}
+
+
+export interface TimeSeriesDisplaySubplot {
+  xAxis?: TimeSeriesDisplaySubplotItemAxis;
+  items: TimeSeriesDisplaySubplotItem[];
+}
```
amarkensteijn, line 39: Should both be updated in the `fews-pi-request` library.

vanmeeuwen: is requested in https://issuetracker.deltares.nl/browse/FEWS-29590 and https://github.com/Deltares/fews-wms-requests/issues/21

src/lib/ChartConfig/timeSeriesDisplayToChartConfig.ts
vanmeeuwen: will be retrieved from the backend soon

PR #316, opened 2023-10-06
315 vue3 authentication and authorization

wkramer:


src/components/map/MapComponent.vue
```diff
@@ -18,10 +19,29 @@
...
 import { configManager } from '../../services/application-config'
+import type { ResourceType } from 'mapbox-gl'
+import { authenticationManager } from '@/services/authentication/AuthenticationManager'
 
 const accessToken = ref('')
 
 onBeforeMount(() => {
   accessToken.value = configManager.get('VITE_MAPBOX_TOKEN')
 })
+
```
wkramer, line 30: Return `{ url }` instead of `{ url: url }`

src/lib/requests/transformRequest.ts
```diff
@@ -0,0 +1,9 @@
+import { authenticationManager } from '@/services/authentication/AuthenticationManager.ts'
+
+export function transformRequestFn(controller?: AbortController) {
```
wkramer, line 3: Rename to `createTransformRequestFn`

PR #345, opened 2023-10-19
Add splash screen and adjust z indices

### LINKED ISSUE #333: Add splash screen
When starting the viewer, one has to explicitly accept the terms and conditions.
splashscreen_v20231009

hvangeffen:
closes #333


src/components/SplashScreen.vue
```diff
@@ -0,0 +1,15 @@
+<template>
+  <v-dialog v-model="showDialog" max-width="900" @click:outside="showDialog = false">
```
vanmeeuwen, line 2: This should be an accept/consent button (on the terms and conditions), not click outside.

src/components/SplashScreen.vue
```diff
@@ -0,0 +1,15 @@
...
+  </v-dialog>
+</template>
+
+<script lang="ts">
+import { Component, Vue } from 'vue-property-decorator'
+
+@Component
+export default class SplashScreen extends Vue {
+  showDialog: boolean = true
+  splashUrl: string = `${process.env.BASE_URL}splash.png`
```
vanmeeuwen, line 13: put it in an images folder

src/views/MetocDataView.vue
```diff
@@ -1,5 +1,6 @@
 <template>
   <div class="display-container">
+    <SplashScreen />
```
vanmeeuwen, line 3: We now have the splashscreen only here, maybe we can skip the base route and always start here as well when we route to the viewer. Otherwise it should also be added in the start layout.

vanmeeuwen (approved):

PR #359, opened 2023-10-24
323 vue3 topology display

amarkensteijn:


ceesvoesenek (approved):
Topology display looks good, also some nice other improvements.

Just some comments about small things

PR #409, opened 2023-11-01
Sync version between cogmenu and splashscreen

### LINKED ISSUE #369: Match cogmenu version to splashscreen image version

hvangeffen:
Closes #369 
image
image

Both use the `package.json` version number


vanmeeuwen (commented):
I notice that the mouse pointer gives the link indication when hovering over the full splash screen. This may be confusing and the Terms of Use doesn't show more to be a link than the rest of the screen.

vanmeeuwen (approved):

PR #421, opened 2023-11-03
Add vite-plugin-vuetify

wkramer:


wkramer:
Will be implemented in https://github.com/Deltares/fews-web-oc/pull/980

PR #473, opened 2023-11-13
469 2x 4x without cache

### LINKED ISSUE #469: Add 2x and 4x speed without caching

hvangeffen:
Closes #469

### DIFF
--- a/src/components/DateTimeSlider.vue
+++ b/src/components/DateTimeSlider.vue
@@ -17,24 +17,24 @@
               mdi-skip-previous
             </v-icon>
           </v-btn>
-          <v-btn :color="playColor" icon @click="togglePlay" ref="PlayButton">
+          <v-btn :color="playColor(speed.base)" icon @click="togglePlay(speed.base)" ref="PlayButton">
             <v-icon>
-              {{ isPlaying ? 'mdi-pause' : 'mdi-play' }}
+              {{ isPlayingAtSpeed(speed.base) ? 'mdi-pause' : 'mdi-play' }}
+            </v-icon>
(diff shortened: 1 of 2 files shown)


vanmeeuwen (commented):
Currently the implementation is that you need to press play and toggle a speed. It is more elegant to press either "play" or "double speed play" or "quadruple speed play".

src/components/DateTimeSlider.vue
```diff
@@ -27,6 +27,16 @@
...
+          <v-btn :color="speedColor(speed.twoTimes)" icon @click="toggleSpeed(speed.twoTimes)" ref="TwoTimesButton">
+            <v-icon>
+              mdi-chevron-double-right
+            </v-icon>
+          </v-btn>
+          <v-btn :color="speedColor(speed.fourTimes)" icon @click="toggleSpeed(speed.fourTimes)" ref="FourTimesButton">
+            <v-icon>
+              mdi-chevron-triple-right
+            </v-icon>
+          </v-btn>
```
vanmeeuwen, line 39: Add these buttons immediately next to the play button.

vanmeeuwen (commented):
We had replaced this.stopPlay() with this.index = 0
Were is this change now? I know it is not part of the current task, but I wonder what happened.

hvangeffen:
> We had replaced this.stopPlay() with this.index = 0 Were is this change now? I know it is not part of the current task, but I wonder what happened.

This change is from a branch which I worked on previously, as such the new code has not been pulled in yet and since there are no conflicts merging is allowed.

PR #549, opened 2023-12-14
add Coastline

UAEpro:
~~we should also add a setting to enable or disable coastline~~
edit: coastline has been moved to it's own layer file


UAEpro:
coastline has been removed from AnimatedMapboxLayer and has been added as a wms file and can be called from SpatialDisplay or any other Displays if needed

PR #575, opened 2023-12-18
Origin of flag cannot change when editing a timeseries.

ekkelenkamp:
Please review. The flag origin should not be changed.


amarkensteijn (changes requested):
The original version is correct. The requirements state that, if a user changes an existing data value, "the flag is changed from original to corrected".

ekkelenkamp:
Just reviewed with Marc and you are correct. The change can mi ignored.

PR #581, opened 2024-01-02
576 enable adjustment of legend ranges

### LINKED ISSUE #576: Enable adjustment of legend ranges

hvangeffen:
Closes #576 
image


vanmeeuwen (approved):

vanmeeuwen:
After changing layer, if the legend range is adjusted, the adjusted value stays the same. It should be layer dependent.

vanmeeuwen:
On mouseover of the legend, not only the min and max change to pointer, but the rest of the values show the same behavior

PR #582, opened 2024-01-03
547 show timeseries when clicking on location

### LINKED ISSUE #547: Show timeseries when clicking on location

hvangeffen:
Closes #547


vanmeeuwen:
Changing layers doesn't stay on the same location.
For example when starting here: http://localhost:5173/topology/node/map/oceanographic_water_level_d3d/location/obs162 and then moving to the cmems, it closes the window and the location is not selected.

hvangeffen:
> Changing layers doesn't stay on the same location. For example when starting here: http://localhost:5173/topology/node/map/oceanographic_water_level_d3d/location/obs162 and then moving to the cmems, it closes the window and the location is not selected.

This is due to issues with the topology router as mentioned in #584

hvangeffen:
This solution won't work for locations that have the same name but a different post fix:
E49d3_N29d3
E49d3_N29d3_850mb

Example case:
image
image

PR #586, opened 2024-01-04
584 fix topology display router issues

hvangeffen:
Before this change the route first goes to: `/topology/node/:nodeId?` and then to `/topology/node/:nodeId?/map/:layerName?`. This causes the `SpatialDisplay` component to rerender and flicker. 
When directly linking to `TopologySpatialDisplay` we get the same behaviour as in the spatial display route (http://localhost:5174/map/oceanographic_water_level_d3d). 

With this change it also seems to be that the keep-alive is not needed anymore:
```vue
  <router-view v-slot="{ Component }">
    <keep-alive include="SpatialDisplay">
      <component :is="Component" />
    </keep-alive>
  </router-view>

<!-- can just be -->
<router-view />

```
But this could potentially still be useful for switching from the `TopologyTimeSeries` tab back to the `TopologySpatialDisplay` tab.

### DIFF
--- a/src/components/spatialdisplay/SpatialDisplayComponent.vue
+++ b/src/components/spatialdisplay/SpatialDisplayComponent.vue
@@ -130,16 +130,21 @@ const legendTitle = computed(() => {
   return `${selectedLayer.value?.title}${unitString}`
 })
 
-watch(times, () => {
-  const timesValue = times.value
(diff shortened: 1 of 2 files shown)


wkramer:
Implemented by #592

PR #589, opened 2024-01-11
Only use BASE_URL

wkramer:


ekkelenkamp (commented):
For the OIDC looks good to me. 
I'm not sure about the the vite.config.ts. Is the base not used here at all?

wkramer:
`process.env.NODE_ENV` and `process.env.BASE_URL` where never set. 
The base was always '/', what is the default setting anyway.

You can only change the base with the command line option.

PR #620, opened 2024-01-18
619 disable edit in table based on backend config

hvangeffen:


src/components/table/TimeSeriesTable.vue
```diff
@@ -142,6 +145,8 @@ const props = withDefaults(defineProps<Props>(), {
 const emit = defineEmits(['change'])
 
 const store = useFewsPropertiesStore()
+const configStore = useConfigStore()
+const readOnlyMode = ref<boolean>(configStore.general.readonlyMode ?? false)
```
vanmeeuwen, line 149: readonlyMode still needs to be added to the configStore

hvangeffen: Can remove extend due to update

hvangeffen: image
It is there since the update

vanmeeuwen (approved):

PR #634, opened 2024-01-18
Set highlightSelectedLocationOnMap on mounted

wkramer:


vanmeeuwen:
Already implemented by the WatchEffect in commit: https://github.com/Deltares/fews-web-oc/pull/630/commits/b0a04de1a24a5436738aea5ab107506452ada004

PR #636, opened 2024-01-18
Allow setting of banner background from css

hvangeffen:


public/weboc-default-style.css
```diff
@@ -6,6 +6,7 @@
   --contrast-color: black;
   --canvas-fill: rgb(223, 223, 223);
   --weboc-app-bar-bg-color: #080c80;
+  --weboc-app-bar-bg-image: url('/images/blue_sky.jpg');
```
vanmeeuwen, line 9: don't set this as web-oc default. Will probably only be used for the metoc

hvangeffen: Should the image then even be uploaded to the public folder?

hvangeffen: Move to custom stylesheet

PR #638, opened 2024-01-18
Add version to cogmenu

hvangeffen:
image


vanmeeuwen (commented):
Add this not in the settings menu, but in the info menu in the sidebar in the end

PR #673, opened 2024-01-29
Support node with only filterIds

wkramer:
Do not display wms layer and controls.

Test with: FewsWebServices wqps `WebOC demo nodes`

### DIFF
--- a/src/services/useWms/index.ts
+++ b/src/services/useWms/index.ts
@@ -93,16 +93,19 @@ export function useWmsLegend(
   async function loadLegend(): Promise<void> {
     const _layers = toValue(layerName)
     const _useDisplayUnits = toValue(useDisplayUnits)
-    if (_layers === '') return
-    try {
-      legendGraphic.value = await wmsProvider.getLegendGraphic({
-        layers: _layers,
-        // Enable when fews-wms-requests is updated
-        //@ts-ignore
-        useDisplayUnits: _useDisplayUnits,
-      })
-    } catch (error) {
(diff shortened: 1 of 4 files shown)


src/services/useWms/index.ts
```diff
@@ -30,21 +30,24 @@ export function useWmsLayerCapabilities(
...
-        importFromExternalDataSource: false,
-        onlyHeaders: false,
-        forecastCount: 1,
-      })
-      if (capabilities.layers.length > 0) {
-        layerCapabilities.value =
-          capabilities.layers.find((l) => l.name === _layers) ??
-          capabilities.layers[0]
+    if (_layers === '') {
+      layerCapabilities.value = undefined
```
hvangeffen, line 34: Could keep the return here, no need to indent the code for an initial condition check

src/services/useWms/index.ts
```diff
@@ -93,16 +96,19 @@ export function useWmsLegend(
...
-      legendGraphic.value = await wmsProvider.getLegendGraphic({
-        layers: _layers,
-        // Enable when fews-wms-requests is updated
-        //@ts-ignore
-        useDisplayUnits: _useDisplayUnits,
-      })
-    } catch (error) {
-      console.error(error)
+    if (_layers === '') {
+      legendGraphic.value = undefined
```
hvangeffen, line 100: Same here

hvangeffen (approved):
Works well when clicking around

PR #694, opened 2024-02-08
Only show component selection if there are more than 1 component

### LINKED ISSUE #690: Don't display components in the sidebar if there is only one
When there is only a single component present, don't show it in the navigation sidebar. E.g. for Durban, overview is the only view available.
image

hvangeffen:
Closes #690 

Does not fix `durban` since it still has 2 components:

```
  "components" : [ {
    "id" : "topologyDisplay",
    "type" : "TopologyDisplay",
    "title" : "Overview",
    "showLeafNodesAsButtons" : false,
    "defaultPath" : {
      "nodeId" : "viewer_saws_1x1"
    }
  }, {
    "id" : "htmlDisplay",
    "type" : "HtmlDisplay",
    "title" : "Terms of use",
    "path" : "terms-of-use",
    "url" : "terms-of-use.html"
  } ]
```

### DIFF
--- a/src/layouts/DefaultLayout.vue
+++ b/src/layouts/DefaultLayout.vue
@@ -14,7 +14,11 @@
           <v-spacer />
           <login-component v-if="configManager.authenticationIsEnabled" />
         </v-toolbar>
-        <v-menu origin="left" min-width="320">
+        <v-menu
+          origin="left"
+          min-width="320"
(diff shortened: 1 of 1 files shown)


wkramer:
I'll check if we have implemented the option to exclude a component from the navigation menu.

PR #696, opened 2024-02-12
Add animated streamline layer to spatial display

ceesvoesenek:


src/components/wms/AnimatedStreamlineMapboxLayer.vue
```diff
@@ -0,0 +1,130 @@
...
+
+  // Create and add layer.
+  layer = new WMSStreamlineLayer(layerId, options)
+  map.value.addLayer(layer, 'boundary_country_outline')
+  map.value.on('dblclick', onDoubleClick)
+
+  // Make sure we are at the appropriate time, elevation and color scale range
+  // after the visualiser has been initialised.
+  layer.on('load', () => {
+    if (!layer || !props.layerOptions) return
```
amarkensteijn, line 45: Is the `!layer` necessary? How can we do a `layer.on('load', ...)` if `layer` does not exist?

src/components/wms/AnimatedMapboxLayer.vue
```diff
@@ -55,24 +58,37 @@ function getCoordsFromBounds(bounds: LngLatBounds) {
...
-      e.tile !== undefined &&
-      e.isSourceLoaded
-    ) {
-      map.value.setPaintProperty(e.sourceId, 'raster-opacity', 1)
-    }
-  })
-  map.value.on('dblclick', (e) => emit('doubleclick', e))
+  map.value.on('moveend', onMapMove)
+  map.value.on('data', onDataChange)
+  map.value.on('dblclick', onDoubleClick)
```
amarkensteijn, line 85: Nice cleanup :)

src/components/wms/AnimatedStreamlineMapboxLayer.vue
```diff
@@ -0,0 +1,130 @@
...
+  const baseUrlWms = `${baseUrl}/wms`
+  return {
+    baseUrl: baseUrlWms,
+    layer: layerOptions.name,
+    streamlineStyle: StreamlineStyle.ColoredParticles,
+    numParticles: streamlineOptions?.numberOfParticles ?? 1000,
+    particleSize: streamlineOptions?.particleSize ?? 3,
+    speedFactor: streamlineOptions?.speedFactor ?? 0.2,
+    fadeAmountPerSecond: streamlineOptions?.fadeAmount ?? 0.1,
+    particleColor: `#${streamlineOptions?.particleColor}`,
```
amarkensteijn, line 120: Wouldn't it be better to set the '#' in the `particleColor`? That way, a `particleColor` of `rgb(r,g,b)` or `'black'` would also be possible.

src/lib/streamlines/types.ts
```diff
@@ -0,0 +1,4 @@
+export enum LayerType {
+  Streamline = 'streamline',
+  Static = 'static',
+}
```
amarkensteijn, line 4: If this is a type, shouldn't we do:
```suggestion
export const LayerTypeEnum = {
  Streamline: 'streamline',
  Static: 'static',
} as const

export type LayerType = (typeof LayerTypeEnum)[keyof typeof LayerTypeEnum]
```

amarkensteijn (changes requested):
Currently, there is still an error for some layers. For instance for http://localhost:5173/topology/node/waves/oceanographic_waves_swan/map/oceanographic_waves_swan, when switching to 'animated', it give an "Uncaught (in promise) RangeError: offset is outside the bounds of the DataView". When going to http://localhost:5173/topology/node/currents/oceanographic_currents_cmems/map/oceanographic_currents_cmems with 'CMEMS' and 'Animated' first, and then switching to 'Oceanographic' -> 'Waves', you do get a animated layer for Swan. Something to do with the time (by comparing the two requests).

src/components/wms/AnimatedStreamlineMapboxLayer.vue
ceesvoesenek: We can't, but TypeScript doesn't know that...

src/lib/streamlines/types.ts
ceesvoesenek: Renamed to `LayerKind`

src/components/wms/AnimatedStreamlineMapboxLayer.vue
ceesvoesenek: Yes it would be, but this is a back-end problem...

ceesvoesenek:
(Force-)pushed changes, problems appear to be fixed.

amarkensteijn:
The multiple `getMap` request seems gone (or, at least, fewer).

However, when quickly switching between layers, I get an unexpected result:
Start at the Swan Waves animated layer:
image
Then, switch to WW3 and immediately switch back to Swan. I'd expect the exact layer (including times and colorrange etc) as before the switch. However, I get:
image
(This also still seems to trigger a double `getMap` request)

amarkensteijn (commented):
The multiple `getMap` request seems gone (or, at least, fewer).

However, when quickly switching between layers, I get an unexpected result:
Start at the Swan Waves animated layer:
image
Then, switch to WW3 and immediately switch back to Swan. I'd expect the exact layer (including times and colorrange etc) as before the switch. However, I get:
image
(This also still seems to trigger a double `getMap` request)

amarkensteijn (approved):

PR #697, opened 2024-02-12
667 maplibre

wkramer:


wkramer:
Remove VITE_MAPBOX_TOKEN from Application Config

PR #700, opened 2024-02-14
653 sidebar should highlight the selected layer

vanmeeuwen:


src/views/TopologyDisplayView.vue
```diff
@@ -182,13 +182,15 @@ watch(
   () => props.nodeId,
   () => {
     if (props.nodeId) {
-      if (
-        typeof props.nodeId === 'string' &&
-        active.value[0] !== props.nodeId
+      if (typeof props.nodeId === 'string' && active.value !== props.nodeId) {
+        console.log('props.nodeid 0', props.nodeId)
```
hvangeffen: Remove console logs

src/views/TopologyDisplayView.vue
```diff
@@ -182,13 +182,15 @@ watch(
...
+        active.value = props.nodeId
+      } else if (
+        Array.isArray(props.nodeId) &&
+        active.value !== props.nodeId[0]
       ) {
-        active.value = [props.nodeId]
-      } else if (active.value[0] !== props.nodeId[0]) {
-        active.value = [props.nodeId[0]]
+        console.log('props.nodeid 1', props.nodeId[0])
```
hvangeffen: Same here

src/services/useMenuItemsStack/index.ts
```diff
@@ -0,0 +1,45 @@
...
+    const _menuItems = toValue(menuItems)
+    const _activeItem = toValue(activeItem)
+
+    if (_activeItem !== '' && _menuItems.length > 0) {
+      const root: ColumnItem = {
+        id: 'rootNode',
+        name: '',
+        children: _menuItems,
+      }
+      const s = [root]
```
hvangeffen, line 22: Could consider:
```suggestion
      stack.value = recursiveFind(root, _activeItem)
    }
  }
  return stack
}

function recursiveFind(item: ColumnItem, id: string): ColumnItem[] {
  if (item.id === id) return [item]

  for (const child of item.children ?? []) {
    const res = recursiveFind(child, id)
    if (res.length > 0) {
      return [item, ...res]
    }
  }

  return []
}

```

hvangeffen (approved):

src/services/useMenuItemsStack/index.ts
vanmeeuwen: Behavior in the column menu is slightly different, so I won't adopt this.

PR #709, opened 2024-02-15
FEWS-30613 PI REST - topology/nodes? - locationId & plotId

grijzea:


ekkelenkamp (changes requested):
locationIds should also be available if plotId is set. Please extend the check with this.

grijzea (commented):
review comments processes

grijzea:
merge after approval from Rudie

PR #710, opened 2024-02-15
707 front end information panel wms layer

martinapippi:


wkramer:
Can we render the colour bars in the panel for each style?

src/components/wms/InformationPanel.vue
```diff
@@ -0,0 +1,204 @@
...
+
+const props = withDefaults(defineProps<Props>(), {
+  layerTitle: '',
+  currentTime: null,
+  forecastTime: null,
+  styles: null,
+  completelyMissing: null,
+  firstValueTime: null,
+  lastValueTime: null,
+  colorScaleRange: null,
```
hvangeffen, line 108: I would make most of these like `currentTime?: Date` such that undefined is the missing data value.
Then we don't have to give defaults below for these and we can just pass values in the rest of the code instead of doing:
```ts
      :completelyMissing="props.layerCapabilities?.completelyMissing ?? null"
```
We can then do:
```ts
      :completelyMissing="props.layerCapabilities?.completelyMissing"
```

This is for most of the places using `null`.

src/components/wms/InformationPanel.vue
```diff
@@ -0,0 +1,204 @@
...
+})
+
+const emit = defineEmits(['style-click', 'color-scale-range-change'])
+const layersIcon = 'mdi-layers'
+const timeIcon = 'mdi-clock-time-four-outline'
+const rangeIcon = 'mdi-layers-edit'
+const colorScalesIcon = 'mdi-palette'
+const selectedStyle = ref<Style | null>(null)
```
hvangeffen, line 116: Same here for example where we can do
```suggestion
const selectedStyle = ref<Style>()
```
And then check for undefined in isSelected.

src/components/spatialdisplay/SpatialDisplayComponent.vue
```diff
@@ -123,16 +143,42 @@ const colorScaleRangeString = computed(() => {
...
+  if (selectedStyle.value !== undefined) {
+    return selectedStyle.value
+  } else if (props.layerCapabilities?.styles !== undefined) {
+    return props.layerCapabilities.styles[0]
+  } else {
+    return null
+  }
+})
+
+const legendGraphic = computed(() => {
```
hvangeffen, line 166: This does not need to be computed, the `loadLegend()` in `useWms` already gives back a `ref` legendGraphic which gets updated in the watchEffect below when `style`'s value changes.

```ts
export function useWmsLegend(
  baseUrl: string,
  layerName: MaybeRefOrGetter<string>,
  useDisplayUnits: MaybeRefOrGetter<boolean>,
  colorScaleRange?: MaybeRefOrGetter<string | undefined>,
  style?: MaybeRefOrGetter<Style | null>,
): Ref<GetLegendGraphicResponse | undefined> {
  ...
  const legendGraphic = ref<GetLegendGraphicResponse>()

  async function loadLegend(): Promise<void> {
    ...
    const _style = toValue(style)
    if (_layers === '') {
      legendGraphic.value = undefined
    } else {
      try {
        legendGraphic.value = await wmsProvider.getLegendGraphic({
           ...
          style: _style?.name,
        })
      } catch (error) {
        console.error(error)
      }
    }
  }

  watchEffect(() => {
    loadLegend()
  })
```

hvangeffen (changes requested):

src/components/wms/InformationPanel.vue
martinapippi: this is fine, except for some of the `props.layerCapabilities` that are sometimes defined as `completelyMissing?: boolean | null` within the `fews-wms-requests`

src/components/spatialdisplay/SpatialDisplayComponent.vue
```diff
@@ -28,6 +28,20 @@
       :locationId="props.locationId"
       @changeLocationId="onLocationChange"
     />
+    <InformationPanel
+      :layerTitle="props.layerCapabilities?.title"
+      :currentTime="currentTime"
+      :forecastTime="forecastTime"
+      :styles="styles"
+      :completelyMissing="
+        props.layerCapabilities?.completelyMissing ? true : false
```
hvangeffen, line 37: This will probably be wrong now if completelyMissing is defined but set to false.
Could be `?? false`

src/components/wms/InformationPanel.vue
```diff
@@ -0,0 +1,201 @@
...
+  if (!props.firstValueTime || !props.lastValueTime) return ''
+  const format = 'dd/MM/yyyy, HH:mm:ss'
+  return `${DateTime.fromJSDate(props.firstValueTime).toFormat(
+    format,
+  )} → ${DateTime.fromJSDate(props.lastValueTime).toFormat(format)}`
+})
+const formattedCurrentTime = computed(() => {
+  if (!props.currentTime) return ''
+  const format = 'HH:mm ZZZZ'
+  const timeZone = 'Europe/Amsterdam'
```
hvangeffen, line 149: Do we always want nl / amsterdam here?
WebOC is used quite broadly, maybe get this from the users locale.

hvangeffen (changes requested):

src/components/wms/InformationPanel.vue
```diff
@@ -0,0 +1,201 @@
...
+  z-index: 1000;
+  top: 8px;
+  right: 10px;
+  border-radius: 5px;
+  backdrop-filter: blur(5px);
+  background-color: rgba(var(--v-theme-surface), 0.8);
+  box-shadow: 0 0 5px rgba(0, 0, 0, 0.5);
+}
+
+.info-icon {
```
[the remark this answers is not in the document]
hvangeffen: Also the other classes below this one

hvangeffen (approved):

PR #727, opened 2024-02-26
DWO-655 Use display/system units setting in WMS layers

ceesvoesenek:
This fixes the difference between animated streamline colours and static WMS colours.

An issue remains that this option is not passed properly from the WebOC-settings to the `SpatialDisplay`, but that is out of scope here.


src/lib/streamlines/layer.ts
```diff
@@ -19,13 +19,13 @@ import {
...
   layer: string
+  style?: string
+  useDisplayUnits?: boolean
   streamlineStyle: StreamlineStyle
   numParticles: number
   particleSize: number
   speedFactor: number
   fadeAmountPerSecond: number
-  style?: string
-  colorScaleRange?: string
```
amarkensteijn, line 28: Shouldn't `colorScaleRange` be a part of the `WMSStreamlineLayerOptions`. I know it is set later (or at least can be set later) through `setColorScaleRange`. But it feel counterintuitive that `colorScaleRange` is part of the `MapboxLayerOptions`, that the streamline layer actually does something with `colorScaleRange`, but that `colorScaleRange` is not part of the `WMSStreamlineLayerOptions`.

ceesvoesenek: I consider it to be similar to `elevation` and `time` (also absent from these options): things that will be set by controls during interaction with the WMS layer.

It is set during initialisation of the layer, and will then indeed be set with `setColorScaleRange`, similar to time and elevation.

PR #732, opened 2024-02-27
Store colour range changes by user

wkramer:


wkramer:
We should store the edited value range per legend style class in a Pinia store.
When switching style class we should check if available in store otherwise use range value from GetLegendGraphic response. We should __not__ retain the current range value when switching legend style class.

PR #740, opened 2024-02-28
fixed bug in helpMenu

martinapippi:


src/layouts/DefaultLayout.vue
```diff
@@ -233,6 +235,12 @@ const versionString = computed(() => {
   return packageConfig.version
 })
 
+const helpMenu = computed(() => {
+  if ('helpMenu' in (configStore.general as any)) {
+    return (configStore.general as any).helpMenu
+  }
+})
```
hvangeffen, line 242: Maybe add a TODO to type this once fews-pi-requests is updated

src/layouts/DefaultLayout.vue
```diff
@@ -70,14 +70,16 @@
...
                   append-icon="mdi-open-in-new"
                   >{{ item.name }}</v-list-item
                 >
                 <v-list-item
-                  v-for="item in (configStore.general as any).helpMenu.path"
+                  v-if="helpMenu"
+                  v-for="item in helpMenu.path"
                   :to="{ name: 'HtmlDisplay', params: { path: item.path } }"
                   href="#"
                   >{{ item.name }}</v-list-item
```
hvangeffen, line 85: Just a note (seperate from this issue), but this currently always orders the urls first followed by the paths.
This would have to be changed in the backend though:
image
Where the items should be able to be ordered interleaved with a type param specifying the type.

hvangeffen (approved):

src/layouts/DefaultLayout.vue
martinapippi: that would be nice but you would still need to have 2 `v-list-item` because you can't specify both `href` and `to` attributes.

hvangeffen: Yes, the v-for would then probably be on a `<template>` parent

PR #760, opened 2024-03-12
Add uploading of dist to assets of newly created releases

hvangeffen:


.github/workflows/release.yml
```diff
@@ -0,0 +1,36 @@
...
+    types:
+      - created
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        node-version: [16, 18.3]
```
wkramer, line 14: Remove matrix and only use single node version (v20)

PR #793, opened 2024-04-04
filter showInNavigationMenu on activeComponents and show 1 element directly if it is the only one

martinapippi:
Closes https://issuetracker.deltares.nl/browse/DWO-657?filter=-1 and https://issuetracker.deltares.nl/browse/DWO-682?filter=-1

### DIFF
--- a/src/layouts/DefaultLayout.vue
+++ b/src/layouts/DefaultLayout.vue
@@ -38,6 +38,10 @@
           <v-list density="compact">
             <v-list-subheader>Switch to</v-list-subheader>
             <v-list-item
+              aria-label="Menu button"
+              class="ma-1"
+              rounded
+              variant="tonal"
               v-for="(item, i) in navigationMenuComponents"
               :key="i"
               :value="item"
@@ -52,6 +56,7 @@
         </v-menu>
         <v-list density="compact" v-if="navigationMenuComponents.length === 1">
           <v-list-item
+            v-if="$router.hasRoute(navigationMenuComponents[0].to.name)"
             :value="navigationMenuComponents[0]"
             :to="navigationMenuComponents[0].to"
           >
--- a/src/layouts/DefaultLayout.vue
+++ b/src/layouts/DefaultLayout.vue
@@ -38,10 +38,6 @@
           <v-list density="compact">
             <v-list-subheader>Switch to</v-list-subheader>
             <v-list-item
-              aria-label="Menu button"
-              class="ma-1"
-              rounded
-              variant="tonal"
               v-for="(item, i) in navigationMenuComponents"
               :key="i"
               :value="item"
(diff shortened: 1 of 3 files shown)


src/layouts/DefaultLayout.vue
```diff
@@ -50,6 +50,19 @@
             </v-list-item>
           </v-list>
         </v-menu>
+        <v-list density="compact" v-if="navigationMenuComponents.length === 1">
+          <v-list-item
+            :value="navigationMenuComponents[0]"
+            :to="navigationMenuComponents[0].to"
```
hvangeffen, line 56: This errors as this is now immediately created on page load for a dynamic route that has not been setup yet: 
image

As such we need to check if the route exists:
```vue
<v-list-item
    v-if="router.hasRoute(navigationMenuComponents[0].to.name)"
```

src/views/AboutView.vue
```diff
@@ -3,7 +3,8 @@
     <v-card-title class="justify-center">
       {{ configStore.general.title ?? 'Delft-FEWS Web OC' }}
     </v-card-title>
-    <v-card-text v-if="configStore.activeComponents.length > 0">
+    <v-card-text v-if="configStore.activeComponents.length > 0"> </v-card-text>
+    <v-card-text v-else-if="configStore.activeComponents.length > 1">
       Select one of the following options to get started.
     </v-card-text>
     <v-card-text v-else>
```
hvangeffen, line 10: Empty v-card-text seems weird here, could make it:
```suggestion
    <v-card-text v-else-if="configStore.activeComponents.length > 1">
      Select one of the following options to get started.
    </v-card-text>
    <v-card-text v-else-if="configStore.activeComponents.length === 0">
```

src/views/AboutView.vue
```diff
@@ -12,6 +13,7 @@
     </v-card-text>
     <v-list density="compact">
       <v-list-item
+        v-if="configStore.activeComponents.length > 1"
```
hvangeffen, line 16: This if should probably go on the `v-list`

src/layouts/DefaultLayout.vue
```diff
@@ -50,6 +50,19 @@
             </v-list-item>
           </v-list>
         </v-menu>
+        <v-list density="compact" v-if="navigationMenuComponents.length === 1">
```
hvangeffen, line 53: Should this still be a `v-list` and should it look different than the style used for the menu button above?:
```vue
            <v-list-item
              aria-label="Menu button"
              v-bind="props"
              class="ma-1"
              rounded
              variant="tonal"
            >
```

hvangeffen:
Also the pptx mentions:
image

Now the v-list only shows on single component, shouldn't there be a `HtmlDisplay` check?

src/views/AboutView.vue
martinapippi: if the configStore.activeComponents.length is <= 1 then nothing should be shown, so also if it is === 1

src/layouts/DefaultLayout.vue
martinapippi: I assumed so, yes

hvangeffen (approved):

PR #807, opened 2024-04-16
Only retrieve the child locations and the required attributes

grijzea:


src/components/download/DataDownloadDisplayComponent.vue
```diff
@@ -426,19 +430,18 @@ function getAttributeValues(locations: Location[]): string[][] {
   const configuredAttributeIds =
     props.topologyNode.dataDownloadDisplay?.attributes.map((item) => item.id)
   configuredAttributeIds?.forEach((item) => attributeValuesMap.push([]))
+  if (configuredAttributeIds == undefined) return attributeValuesMap
```
hvangeffen, line 433:
```suggestion
  if (configuredAttributeIds === undefined) return attributeValuesMap
```

PR #810, opened 2024-04-17
Allow locations icons to be SVG

ceesvoesenek:


ceesvoesenek:
This is waiting for a back-end change before we can test it properly. SVG icons are now always returned with the icon name `buffered`, which does not exist.

ceesvoesenek:
The backend has been fixed, so this PR is ready for review.

PR #840, opened 2024-04-24
fixed some small bugs

grijzea:
added qualifier to parameter list

### DIFF
--- a/src/components/download/DataDownloadDisplayComponent.vue
+++ b/src/components/download/DataDownloadDisplayComponent.vue
@@ -161,7 +161,7 @@ import { ParameterQualifiersHeader } from '@/lib/download/types'
 import { isEqual } from 'lodash-es'
 import { filterToParams } from '@/lib/download/downloadFiles.ts'
 import { DataDownloadFilter } from '@/lib/download/types/DataDownloadFilter.ts'
-import {uniqWith} from "lodash";
+import { uniqWith } from 'lodash'
 
 interface Props {
   nodeId?: string | string[]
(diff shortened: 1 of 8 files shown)


src/components/download/DataDownloadDisplayComponent.vue
```diff
@@ -185,18 +189,35 @@ const filterId = props.topologyNode.filterIds
...
+const parameterQualifiers = ref<ParameterQualifiersHeader[]>([])
+const selectedParameterQualifiers = ref<ParameterQualifiersHeader[]>([])
 
 allLocations.value = await getLocations()
 locations.value = allLocations.value
 selectedLocations.value = allLocations.value
 
-parameters.value = await getParameters()
-selectedParameters.value = parameters.value
+const timeSeriesResults = await getTimeSeriesHeaders()
```
ceesvoesenek, line 203: Top-level `await` blocks setting up the component until the request finishes, is this what we want?

In any case, since the fetches are independent, they can be executed concurrently with `Promise.all` or `Promise.allSettled`.

src/components/download/DataDownloadDisplayComponent.vue
```diff
@@ -185,18 +189,35 @@ const filterId = props.topologyNode.filterIds
...
+allParameters.value = await getParameters()
+const parameterQualifiersHeaders: ParameterQualifiersHeader[] = []
+timeSeriesResults?.forEach((timeSeriesResult) => {
```
ceesvoesenek, line 206: Could be simplified with a `.map` and a lodash `uniqWith` maybe

src/components/download/DataDownloadDisplayComponent.vue
```diff
@@ -387,15 +419,46 @@ function getLocationName(location: Location): string {
   return location.locationId
 }
 
-function getParameterName(parameter: TimeSeriesParameter): string {
-  const showParameterName =
-    props.topologyNode.dataDownloadDisplay?.showParameterName
+function getParameterName(
+  parameters: TimeSeriesParameter[],
+  parameterQualifiersHeader: ParameterQualifiersHeader,
+  showParameterName: string | undefined,
```
ceesvoesenek, line 425: Maybe `showParameterName?: 'name' | 'short name'`

src/lib/download/downloadFiles.ts
```diff
@@ -35,6 +35,16 @@ async function downloadFileWithFetch(
   }
 }
 
+export function filterToParams(filter: Record<string, any>): string {
```
ceesvoesenek, line 38: If we're making a new version of `filterToParams`, we could also consider using [the URL API for query parameters](https://developer.mozilla.org/en-US/docs/Web/API/URL/searchParams), which might be a bit nicer to work with than strings?

src/lib/download/types/DataDownloadFilter.ts
```diff
@@ -0,0 +1,6 @@
+export interface DataDownloadFilter {
```
ceesvoesenek, line 1: Should this extend a `TimeSeriesFilter` (or something), since we are using it on the FEWS PI time series end point?

src/lib/download/types/ParameterQualifiersHeader.ts
```diff
@@ -0,0 +1,4 @@
+export interface ParameterQualifiersHeader {
```
ceesvoesenek, line 1: Does `@deltares/fews-pi-requests` not have an equivalent type? Maybe you can use [`Pick`](https://www.typescriptlang.org/docs/handbook/utility-types.html#picktype-keys) to extract a subset of this type, to guarantee that the types remain consistent.

PR #849, opened 2024-04-26
Always zoom to WMS bounding box when switching layers

### LINKED ISSUE #848: Always zoom to WMS bounding box when switching nodes
https://issuetracker.deltares.nl/browse/DWO-718

hvangeffen:
Closes #848


wkramer:
Does not seem to work when selection node with button
image

PR #877, opened 2024-05-29
Set filename based on content disposition

hvangeffen:


hvangeffen:
Backend needs to have: `'Access-Control-Expose-Headers', 'Content-Disposition'`

PR #884, opened 2024-05-29
BoundingBox: improved the string conversion to ensure rounding

MatTolladay:
This doesn't guarantee that the individual members of the BoundingBox will display correctly when accessed.

### DIFF
--- a/src/components/map/BoundingBoxControl.vue
+++ b/src/components/map/BoundingBoxControl.vue
@@ -19,22 +19,15 @@
 </template>
 
 <script setup lang="ts">
-import { computed } from 'vue'
-
-import { BoundingBox, boundingBoxToString } from '@/services/useBoundingBox'
+import { useBoundingBox } from '@/services/useBoundingBox'
 
 import DrawBoundingBoxControl from '@/components/map/DrawBoundingBoxControl.vue'
 
-const boundingBox = defineModel<BoundingBox | null>('boundingBox', {
-  default: null,
-})
+const { boundingBox, boundingBoxString } = useBoundingBox(1, 1)
+
 const isActive = defineModel<boolean>('active', { default: false })
 
 const emit = defineEmits(['finish'])
-
-const boundingBoxString = computed(() =>
-  boundingBox.value ? boundingBoxToString(boundingBox.value) : '—',
-)
 
 function onFinish(): void {
   isActive.value = false
--- a/src/components/map/BoundingBoxControl.vue
+++ b/src/components/map/BoundingBoxControl.vue
@@ -19,15 +19,20 @@
 </template>
 
 <script setup lang="ts">
-import { useBoundingBox } from '@/services/useBoundingBox'
-
+import { computed } from 'vue'
+import { BoundingBox, boundingBoxToString } from '@/services/useBoundingBox'
 import DrawBoundingBoxControl from '@/components/map/DrawBoundingBoxControl.vue'
 
-const { boundingBox, boundingBoxString } = useBoundingBox(1, 1)
-
+const boundingBox = defineModel<BoundingBox | null>('boundingBox', {
+  default: null,
+})
 const isActive = defineModel<boolean>('active', { default: false })
 
 const emit = defineEmits(['finish'])
+
+const boundingBoxString = computed(() =>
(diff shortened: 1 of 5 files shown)


src/components/map/BoundingBoxControl.vue
```diff
@@ -19,23 +19,16 @@
...
-import { BoundingBox, boundingBoxToString } from '@/services/useBoundingBox'
+import { useBoundingBox } from '@/services/useBoundingBox'
 
 import DrawBoundingBoxControl from '@/components/map/DrawBoundingBoxControl.vue'
 
-const boundingBox = defineModel<BoundingBox | null>('boundingBox', {
-  default: null,
-})
+const { boundingBox, boundingBoxString } = useBoundingBox(1, 1)
+
```
hvangeffen, line 27: This creates a new bounding box state object, disconnecting the 
image
bbox state and the 
image
bbox state.

The import used before was accessing global state.

src/services/useBoundingBox/index.ts
```diff
@@ -52,8 +53,25 @@ function roundBoundingBox(
...
-export function boundingBoxToString(boundingBox: BoundingBox): string {
-  return `${boundingBox.lonMin}°E ${boundingBox.latMin}°N, ${boundingBox.lonMax}°E ${boundingBox.latMax}°N`
+export function boundingBoxToString(
+  boundingBox: BoundingBox,
+  longitudeStepSize: number,
+  latitudeStepSize: number,
+): string {
+  const asPrecision = (value: number, step: number): string => {
+    const nIntegers = Math.floor(Math.log10(Math.abs(value)))
+    const stepStrs = step !== undefined ? step.toString().split('.') : '1'
```
hvangeffen, line 63: since asPrecision only excepts a `:number` step can not be undefined in this case

src/services/useBoundingBox/index.ts
```diff
@@ -52,8 +53,25 @@ function roundBoundingBox(
...
+    const nDecimalPlaces = stepStrs.length > 1 ? stepStrs[1].length : 2
+    const nPrecision = Math.round(
+      Math.min(Math.max(nIntegers + nDecimalPlaces, 2), 99),
+    )
+    return value.toPrecision(nPrecision)
```
hvangeffen, line 68: would [toFixed(nDecimalPlaces)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toFixed) be the same?

src/components/map/BoundingBoxControl.vue
MatTolladay: I have reverted the changes. I don't understand how the old import accessed global state, but I would like to know. I think I understand how the `const { boundingBox, boundingBoxString } = useBoundingBox(1, 1)` creates a new box (the `_boundingBox` instantiated in the `useBoundingBox` function. Is that correct?

hvangeffen: Not global state through import sorry,

Definemodel (a compiler macro for prop / emit definition) was accepting the bbox through a prop:
https://vuejs.org/guide/components/v-model.html

MatTolladay: Ahh, the boundingBox is being provided by the parent component as a `v-model`, using the `workflowsStore.boundingBox`. This does not provide reactivity to the lon/lat values in `BoundingBoxControl` when selecting a box.

hvangeffen:
For the form itself maybe we can round the number after selecting a box with the draw tool:
image

src/lib/utils/math.ts
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * Round a number to the nearest step
+ * @param {number} value - The number to round.
+ * @param {number} step - The step to round to.
+ * @returns {number} The rounded number.
+ * @example
+ * roundToStep(54.6545, 0.1) // 54.7
+ */
+export function roundToStep(value: number, step: number): number {
```
ceesvoesenek, line 27: What if the step is 0.2? It will round to 0.1, right? Is that intended behaviour? If so, maybe this should be called `roundToStepPrecision` or something?

hvangeffen: Renamed to roundToStepPrecision

PR #885, opened 2024-05-29
Added an MGL scalebar to the map

MatTolladay:


hvangeffen:
@wkramer do we want to be able to toggle the scale from the menu?

PR #903, opened 2024-06-10
Get fileName from topology

hvangeffen:


src/lib/download/downloadFiles.ts
```diff
@@ -70,7 +70,7 @@ export async function downloadFileAttachment(
 
 export async function downloadFileWithXhr(
   url: string,
-  extension?: string,
+  fileName?: string,
```
ceesvoesenek, line 73: It's not strictly the filename as it is appended to a timestamp? Maybe something like `filenameSuffix`?

src/stores/workflows.ts
```diff
@@ -48,6 +48,7 @@ const useWorkflowsStore = defineStore('workflows', {
       type: WorkflowType,
       filter: PartialRunTaskFilter | PartialProcessDataFilter,
       body?: string,
```
ceesvoesenek, line 50: I'm not a big fan of having to pass `undefined` arguments to use a single optional argument later in the list. Maybe pass an options object with optional `body` and `fileName` (or `fileNameSuffix`..) fields?

ceesvoesenek (commented):
I don't have so much context on what this is supposed to do, but I trust that that is fine.

Some small comments only

src/stores/workflows.ts
hvangeffen: I agree, I wanted named params from Python at this moment

src/lib/download/downloadFiles.ts
hvangeffen: Changed the name

src/stores/workflows.ts
hvangeffen: Added to an options object

ceesvoesenek (approved):

PR #904, opened 2024-06-11
Use period from display configuration as x-axis domain

ceesvoesenek:
Previously, the domain of the x-axis was dependent on the data, which meant that different charts in the same display could have different x-axis domains. We now use the `period` from a `DisplayConfig` to set the x-axis range instead.

### DIFF
--- a/src/lib/charts/timeSeriesDisplayToChartConfig.ts
+++ b/src/lib/charts/timeSeriesDisplayToChartConfig.ts
@@ -1,6 +1,7 @@
 import type { ChartConfig } from './types/ChartConfig.js'
 import type { ChartSeries } from './types/ChartSeries.js'
 import type {
+  ActionPeriod,
   TimeSeriesDisplayPlotItemXAxis,
   TimeSeriesDisplaySubplot,
   TimeSeriesDisplaySubplotItem,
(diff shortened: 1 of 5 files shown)


src/lib/charts/timeSeriesDisplayToChartConfig.ts
```diff
@@ -16,14 +17,33 @@ import {
...
+    const domain: [Date, Date] = [
+      convertFewsPiDateTimeToJsDate(period.startDate, timeZoneOffsetString),
+      convertFewsPiDateTimeToJsDate(period.endDate, timeZoneOffsetString),
+    ]
+    // Set the domain of the x-axis to be equal to the period, even if there are
+    // no data points in parts of the period.
+    if (xAxis.length === 0) {
+      xAxis.push({ domain })
+    } else {
+      xAxis[0].domain = domain
```
hvangeffen, line 40: Is only the first xAxis relevant?

src/lib/date/convertFewsPiDateTimeToJsDate.test.ts
hvangeffen, line 1: Is the default to have a seperate test file for each function in the date lib?

hvangeffen:
Looks good, was there a test case where previously this had different behaviour?

ceesvoesenek:
> Looks good, was there a test case where previously this had different behaviour?

Yes, many of the locations for the Durban backend show different x-axes for different charts in the same display

hvangeffen:
Seems to work

# Before 
image
# After
image

src/lib/charts/timeSeriesDisplayToChartConfig.ts
```diff
@@ -16,14 +17,33 @@ import {
...
 
 export function timeSeriesDisplayToChartConfig(
   subplot: TimeSeriesDisplaySubplot,
   title: string,
+  period?: ActionPeriod,
 ): ChartConfig {
+  const xAxis = subplot.xAxis ? xAxisFromPlotItemXAxis(subplot.xAxis) : []
+  if (period) {
+    // The period is always specified in UTC.
+    const timeZoneOffsetString = 'Z'
```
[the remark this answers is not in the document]
ceesvoesenek: I wanted to be explicit... I'll remove the default to enforce being explicit about the time zone.

src/lib/charts/timeSeriesDisplayToChartConfig.ts
```diff
@@ -16,14 +17,33 @@ import {
   AxisPosition,
   AxisType,
 } from '@deltares/fews-web-oc-charts'
+import { convertFewsPiDateTimeToJsDate } from '../date/index.js'
```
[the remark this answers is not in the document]
ceesvoesenek: Good point, automatic VSCode import always does this :(

src/lib/charts/timeSeriesDisplayToChartConfig.ts
ceesvoesenek: Yes, for time series there will always be just the one x-axis. I'll add a comment.

ceesvoesenek:
Pushed some changes based on your comments @hvangeffen

PR #905, opened 2024-06-11
Only show workflow button if configured for topology node.

ekkelenkamp:


src/views/TopologyDisplayView.vue
```diff
@@ -7,7 +7,7 @@
       :items="items"
     />
   </Teleport>
-  <Teleport to="#web-oc-toolbar-target">
+  <Teleport to="#web-oc-toolbar-target" v-if="showWorkflowsControl">
     <WorkflowsControl
```
hvangeffen, line 11: Maybe the check should go on the control itself, in case we get more controls here

PR #913, opened 2024-06-12
Show WMS maximum values below the date/time slider

ceesvoesenek:


src/components/general/DateTimeSliderValues.vue
```diff
@@ -0,0 +1,99 @@
...
+    .attr('fill', colourFunc)
+}
+</script>
+
+<style scoped>
+.container {
+  height: 5px;
+  display: flex;
+}
+</style>
```
hvangeffen, line 99: This style is currently unused as there is no class with `.container`

src/components/spatialdisplay/SpatialDisplayComponent.vue
hvangeffen, line 1: Does the `SchematicStatusDisplay` also want this? Don't know of a backend that has it, but there is another DateTimeSlider there.

ceesvoesenek: Not that I know of, if so, maybe that's something for a separate PR as it will have to use a different endpoint.

ceesvoesenek:
> Looks really nice in the viewer, great addition.
> 
> Some layers have coastal artifacts in Durban, which make the bar monocolored. But that is backend. image image

Yeah that's a bummer 😢

src/components/general/DateTimeSliderValues.vue
ceesvoesenek: Good point, I'll remove it

PR #915, opened 2024-06-13
Add ability to set the start and endtime of the graph data

hvangeffen:
Date input currently shows 13/06/2024 as it is formatted according to:
```ts
new Intl.DateTimeFormat(LOCALE_SET_FOR_VUETIFY, { 
    year: 'numeric', month: '2-digit', day: '2-digit'
}).format(date)
```

### DIFF
--- a/src/lib/TimeControl/interval.ts
+++ b/src/lib/TimeControl/interval.ts
@@ -1,4 +1,4 @@
-import { TimeSettingsViewPeriodPreset } from '@deltares/fews-pi-requests'
+import type { TimeSettingsViewPeriodPreset } from '@deltares/fews-pi-requests'
 import type { DurationLikeObject } from 'luxon'
 
 export type Interval = IntervalItem | 'default' | 'custom'
@@ -11,7 +11,7 @@ export interface IntervalItem {
 
 export function periodPresetToIntervalItem(
   preset: TimeSettingsViewPeriodPreset,
-) {
+): IntervalItem {
   const start = {
     [preset.unit]: Number(preset.start),
   }
--- a/src/lib/TimeControl/interval.ts
+++ b/src/lib/TimeControl/interval.ts
@@ -1,0 +1,26 @@
+import { TimeSettingsViewPeriodPreset } from '@deltares/fews-pi-requests'
+import type { DurationLikeObject } from 'luxon'
+
+export type Interval = IntervalItem | 'default' | 'custom'
+
+export interface IntervalItem {
+  label: string
+  start?: DurationLikeObject
+  end?: DurationLikeObject
+}
+
+export function periodPresetToIntervalItem(
+  preset: TimeSettingsViewPeriodPreset,
+) {
+  const start = {
+    [preset.unit]: Number(preset.start),
+  }
+  const end = {
+    [preset.unit]: Number(preset.end),
+  }
+  return {
+    label: preset.label,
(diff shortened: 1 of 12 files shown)


src/lib/TimeControl/interval.ts
```diff
@@ -0,0 +1,26 @@
+import { TimeSettingsViewPeriodPreset } from '@deltares/fews-pi-requests'
```
ceesvoesenek, line 1: This can also be `import type`.

src/services/useDisplayConfig/index.ts
```diff
@@ -10,6 +10,7 @@ import { DisplayConfig } from '../../lib/display/DisplayConfig.js'
 import { timeSeriesDisplayToChartConfig } from '../../lib/charts/timeSeriesDisplayToChartConfig.js'
 import { createTransformRequestFn } from '@/lib/requests/transformRequest.js'
 import { MD5 } from 'crypto-js'
+import { convertFewsPiDateTimeToJsDate } from '@/lib/date/index.js'
```
ceesvoesenek, line 13: Remove the `index.js` (probably auto-import?)

src/services/useDisplayConfig/index.ts
```diff
@@ -41,16 +42,34 @@ function isTimeSeriesGridActionsFilter(
...
+    if (period) {
+      const periodStart = convertFewsPiDateTimeToJsDate(
+        period.startDate,
+        timeZoneOffsetString,
+      )
+      const periodEnd = convertFewsPiDateTimeToJsDate(
+        period.endDate,
+        timeZoneOffsetString,
+      )
+      configPeriod = [startTime ?? periodStart, endTime ?? periodEnd]
```
ceesvoesenek, line 67: We don't need to convert the FEWS PI date/times if we already know we're not going to use them?

src/services/useDisplayConfig/index.ts
```diff
@@ -87,14 +108,29 @@ export function useDisplayConfig(
   const displayConfig = ref<DisplayConfig>()
   const displays = ref<DisplayConfig[]>()
 
+  const response = ref<ActionsResponse>()
+
   watchEffect(async () => {
```
ceesvoesenek, line 113: Maybe add a brief comment on why the fetch is separated from the interpretation of the response?

src/services/useDisplayConfig/index.ts
```diff
@@ -117,24 +153,25 @@ export function useDisplayConfig(
...
-    let nodeId = undefined
     if (isFilterActionsFilter(_filter)) {
       if (!_filter.filterId) return
-      response = await piProvider.getFilterActions(_filter)
+      response.value = await piProvider.getFilterActions(_filter)
     } else if (isTimeSeriesGridActionsFilter(_filter)) {
-      response = await piProvider.getTimeSeriesGridActions(_filter)
-      response.results.forEach((result) => {
+      response.value = await piProvider.getTimeSeriesGridActions(_filter)
+      response.value.results.forEach((result) => {
```
ceesvoesenek, line 174: Not your code, and I haven't looked in too much detail, but could this just be `flatMap`?

ceesvoesenek (commented):
Good work, seems to have been an annoying problem! Thanks for fixing my mess too 😉 

Some comments, but nothing dramatic

ekkelenkamp (approved):
LGTM

PR #932, opened 2024-06-18
Show loading icon when get request for time series is pending

hvangeffen:


src/components/charts/LoadingOverlay.vue
```diff
@@ -0,0 +1,39 @@
...
+    right?: number
+    bottom?: number
+    left?: number
+  }
+}
+
+const props = defineProps<Props>()
+
+const style = computed(() => ({
+  paddingTop: props.margin?.top ? `${props.margin.top}px` : undefined,
```
ceesvoesenek, line 22: It looks a bit weird that you're setting the padding with a property called `margin`. Maybe add a comment why this is the case?

src/components/charts/LoadingOverlay.vue
```diff
@@ -0,0 +1,39 @@
+<template>
+  <div class="loading-overlay" :style="style">
+    <v-icon size="x-large">mdi-loading mdi-spin</v-icon>
```
ceesvoesenek, line 3: There is also `<v-progress-circular indeterminate />` that we already use as a loading icon for the WMS layers, and which has a slightly cooler animation 😄

src/components/charts/TimeSeriesChart.vue
```diff
@@ -1,5 +1,6 @@
 <template>
   <div class="chart-with-chips">
+    <LoadingOverlay v-if="isLoading" class="chart-overlay" :margin="margin" />
```
ceesvoesenek, line 3: This class is no longer used?

src/services/useTimeSeries/index.ts
```diff
@@ -21,6 +21,7 @@ export interface UseTimeSeriesReturn {
   series: Ref<Record<string, Series>>
   isReady: Ref<boolean>
   isLoading: Ref<boolean>
+  loadingRequests: Ref<string[]>
```
ceesvoesenek, line 24: Maybe `loadingSeriesIds` or something? We have quite a lot of different IDs floating around so it's nice to know which one this array contains.

ceesvoesenek (commented):
Looks nice! Some minor comments

src/components/charts/LoadingOverlay.vue
wkramer: Discussed with Tom. Use skeleton loader.

PR #940, opened 2024-06-20
Only allow double click if onlyGrids has been set or not set at all.

ekkelenkamp:


ekkelenkamp:
Over the shoulder review done by Andre.

PR #946, opened 2024-06-20
TopologyDisplayView: The tab type remains the same (if possible) when the user selects a new node.

MatTolladay:
The v-btn-toggle was connecting its v-model to the v-btn keys. This caused the model to become meaningless when the node was changed. Using the tab.type as the button value allows the toggle group model to behave as expected and simplifies the code by removing the need for separate activeTab and activeTabType variables, with activeTab now containing the tab.type string.

Some additional cleaning up was performed

### DIFF
--- a/src/views/TopologyDisplayView.vue
+++ b/src/views/TopologyDisplayView.vue
@@ -73,6 +73,7 @@
       <v-btn
         v-for="tab in displayTabs"
         :key="tab.id"
+        :value="tab.type"
         :href="tab.href"
         :target="tab.target"
         :to="tab.to"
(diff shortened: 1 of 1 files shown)


hvangeffen (commented):
Seems to work nicely.
Only when switching between `dataDownloadDisplay` the state is not updated.
This is because `dataDonwnloadDisplay` is not reactive on `topologyNode` prop. 
Seperate issue though, but could maybe be fixed in: https://github.com/Deltares/fews-web-oc/pull/947, by adding a key on `topologyNode` to the new `dataDownloadDisplayView`

PR #999, opened 2024-09-24
Rescale plot when line is toggled

hvangeffen:
### Description

Redraws the extend when a line is toggled. Currently uses custom add and remove chart functions. Could be added to chart library instead. Resetting extend in case of shared zoom can lead to unexpected behaviour, don't know exactly what would be expected in those cases.

Test example (durban):
http://localhost:5173/topology/node/viewer_coastal_waves/viewer_coastal_waves_ww3/map/ww3_hs/coordinates/-31.159/31.423

### Checklist
    - `rel: new feature`
    - `rel: improvement`
    - `rel: fixes`
    - `rel: ignore`
- [ ] Update documentation.
- [ ] Update tests.

### DIFF
--- a/src/components/charts/TimeSeriesChart.vue
+++ b/src/components/charts/TimeSeriesChart.vue
@@ -47,6 +47,7 @@ import { ref, watch, onMounted, onBeforeUnmount, nextTick } from 'vue'
 import {
   AlertLines,
   CartesianAxesOptions,
+  Chart,
   ChartArea,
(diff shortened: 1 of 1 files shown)


wkramer:
@hvangeffen I think we should fix this in `fews-web-oc-charts`. May be we should have property if a chart should be excluded when determining the data extent for certain axis.

wkramer:
No longer required

PR #1015, opened 2024-09-30
Center selected location marker when location is point geometry

hvangeffen:
### Description

Center selected location marker.

#### Before
image
#### After
image

### Checklist
    - `rel: new feature`
    - `rel: improvement`
    - `rel: fixes`
    - `rel: ignore`
- [ ] Update documentation.
- [ ] Update tests.

### DIFF
--- a/src/components/wms/LocationsLayer.vue
+++ b/src/components/wms/LocationsLayer.vue
@@ -44,7 +44,7 @@ import {
   MglGeoJsonSource,
   MglMarker,
 } from '@indoorequal/vue-maplibre-gl'
-import { FeatureCollection, Geometry } from 'geojson'
+import type { Feature, FeatureCollection, Geometry } from 'geojson'
 import { type Location } from '@deltares/fews-pi-requests'
 import {
   LngLat,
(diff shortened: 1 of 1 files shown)


src/components/wms/LocationsLayer.vue
```diff
@@ -95,11 +95,20 @@ const selectedLocationCoordinates = computed(() => {
   const selectedLocation = geojson.value.features.find(
     (feature) => feature.properties.locationId === props.selectedLocationId,
   )
-  const lat = selectedLocation?.properties.lat
-  const lng = selectedLocation?.properties.lon
-  if (!lat || !lng) return
 
-  return new LngLat(+lng, +lat)
+  if (selectedLocation?.geometry.type == 'Point') {
```
amarkensteijn, line 99: Why not `if (selectedLocation?.geometry.type === Point)` or maybe even something like `if (selectedLocation?.geometry instanceof Point)`?

src/components/wms/LocationsLayer.vue
```diff
@@ -95,11 +95,20 @@ const selectedLocationCoordinates = computed(() => {
...
+    const lng = geometry.coordinates[0]
+    const lat = geometry.coordinates[1]
+
+    return new LngLat(lng, lat)
+  } else {
+    const lat = selectedLocation?.properties.lat
+    const lng = selectedLocation?.properties.lon
+    if (!lat || !lng) return undefined
+
+    return new LngLat(+lng, +lat)
```
amarkensteijn, line 110: Can we move the `return new LngLat(lng, lat)` outside of the if-statements?

hvangeffen: I feel like that will add more lines and make the code slightly less readable.

src/components/wms/LocationsLayer.vue
hvangeffen: Made it a triple === equals

amarkensteijn (approved):

PR #1046, opened 2024-11-07
Fix width of date table column

wkramer:
### Description

The date column in a time series table has a fixed width, giving more space to the value columns.

image


hvangeffen (commented):
The only difference I seem to notice is the Date header being aligned to the top. Removing the added 24ch locally doesn't seem to change the width of the column.

# Old
Screenshot 2024-11-11 at 09-48-53 SMART Polder System
# This PR
Screenshot 2024-11-11 at 09-49-06 SMART Polder System

PR #1054, opened 2024-11-12
Remove report id from report selection and add current subtitle

hvangeffen:
### Description

Makes the reports toolbar the same size as the top bar, removes report id as it was confusing and adds current subtitle to the current report.

### Screenshots

image

### Checklist
    - `rel: new feature`
    - `rel: improvement`
    - `rel: fixes`
    - `rel: ignore`
- [x] Update documentation.
- [x] Update tests.


wkramer:
Current should be selected by default.

PR #1079, opened 2024-11-21
Fix map style change (light to dark) removing added layers

hvangeffen:
### Description

Fixes style changing resulting in the removal of custom layers by adding a prefix to sources and layer ids.

### Checklist
    - `rel: new feature`
    - `rel: improvement`
    - `rel: fixes`
    - `rel: ignore`
- [x] Update documentation.
- [x] Update tests.

### DIFF
--- a/src/components/wms/AnimatedRasterLayer.vue
+++ b/src/components/wms/AnimatedRasterLayer.vue
@@ -3,7 +3,7 @@
 </template>
 
 <script setup lang="ts">
-import { onMounted, onUnmounted, ref, watch } from 'vue'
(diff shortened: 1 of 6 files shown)


src/components/wms/AnimatedRasterLayer.vue
```diff
@@ -38,7 +39,8 @@ const emit = defineEmits(['doubleclick'])
 
 const { map } = useMap()
 
-let currentLayer: string = ''
+const currentLayer = ref('')
```
wkramer, line 42: Why are these ref's?

PR #1136, opened 2024-12-17
Refresh timeseries automatically every 30s

amarkensteijn:
### Description

Performs a timeseries request every 30 seconds if a chart is open, to keep the data up-to-date. The chart is prevented from being redrawn completely, because a redraw would reset the zoom and would create a noticeable/visible 'flicker' of the chart.

### DIFF
--- a/src/components/charts/TimeSeriesChart.vue
+++ b/src/components/charts/TimeSeriesChart.vue
@@ -147,7 +147,7 @@ watch(
   (newValue) => {
     if (axisTime.value) {
       axisTime.value.setDateTime(newValue)
-      onValueChange()
+      axisTime.value.redraw()
     }
   },
 )
--- a/src/components/charts/TimeSeriesChart.vue
+++ b/src/components/charts/TimeSeriesChart.vue
@@ -84,6 +84,7 @@ const legendTags = ref<Tag[]>([])
 const showThresholds = ref(true)
 const chartContainer = ref<HTMLElement>()
 const axisTime = ref<CurrentTime>()
+const hasLoadedOnce = ref(false)
 
 onMounted(() => {
   if (chartContainer.value) {
@@ -129,10 +130,7 @@ onMounted(() => {
     axis.accept(mouseOver)
     axis.accept(axisTime.value)
     resize()
-    if (props.config !== undefined) {
-      refreshChart()
-      setTags()
-    }
+    if (props.config !== undefined) onValueChange()
     window.addEventListener('resize', resize)
   }
 })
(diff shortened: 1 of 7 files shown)


amarkensteijn:
Currently, the charts get recreated every time the data changes, which lead to frequent 'flickering' of the chart (see also [this comment](https://issuetracker.deltares.nl/browse/DWO-816?focusedId=970518&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-970518)). Created [this issue](https://github.com/Deltares/fews-web-oc-charts/issues/120) to look into this.

amarkensteijn:
The interval / automatic refresh still needs to be implemented for the elevation charts

amarkensteijn:
It still flickers sometimes. As far as I've been able to determine, this is due to the `v-show` and the `LoadingOverlay` based on the `isLoading` prop. We should move the loading indicator to each indiviual series, e.g. in the legend tag. We could consider an initial loader indicator on the full chart panel when opening it for the first time, untill the chart config have loaded or untill as least one series has data. Something like this is done for the RWsOS viewer

amarkensteijn:
> The interval / automatic refresh still needs to be implemented for the elevation charts

Has been fixed

amarkensteijn:
> Currently, the charts get recreated every time the data changes, which lead to frequent 'flickering' of the chart (see also [this comment](https://issuetracker.deltares.nl/browse/DWO-816?focusedId=970518&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-970518)). Created [this issue](https://github.com/Deltares/fews-web-oc-charts/issues/120) to look into this.

Has been fixed

amarkensteijn:
> It still flickers sometimes. As far as I've been able to determine, this is due to the `v-show` and the `LoadingOverlay` based on the `isLoading` prop. We should move the loading indicator to each indiviual series, e.g. in the legend tag. We could consider an initial loader indicator on the full chart panel when opening it for the first time, untill the chart config have loaded or untill as least one series has data. Something like this is done for the RWsOS viewer

Has been fixed by removing the loading overlay. It might be nice to have some kind of loading indicator, somewhere, but we should pick that up in a separate issue

hvangeffen (changes requested):
The charts sometimes don't update when switching locations on the map.

amarkensteijn:
> The charts sometimes don't update when switching locations on the map.

Has been fixed by removing the debounced on the loading times series ids

hvangeffen (approved):

PR #1156, opened 2025-01-14
Handle error for location request for topology node with filterIds 

wkramer:
### Description

Correct error handling for locations request for topology node with filterIds. When switching to a node where the locations request failed old locations where still visible on the map.


wkramer:
Requires: https://github.com/Deltares/fews-web-oc-utils/pull/26

PR #1197, opened 2025-02-10
Fix regression causing current location ids to not be saved

hvangeffen:
### Description
Fix current location ids not being saved.
Important for dashboard as it uses it to maintain selected location state.

### Checklist
    - `rel: new feature`
    - `rel: improvement`
    - `rel: fixes`
    - `rel: ignore`
- [x] Update documentation.
- [x] Update tests.

### DIFF
--- a/src/components/spatialdisplay/SpatialDisplay.vue
+++ b/src/components/spatialdisplay/SpatialDisplay.vue
@@ -239,6 +239,7 @@ function openLocationsTimeSeriesDisplay(locationIds: string[]) {
     .replace('WithCoordinates', '')
   currentLatitude.value = undefined
   currentLongitude.value = undefined
+  currentLocationIds.value = locationIds
   router.push({
     name: routeName,
     params: {
--- a/src/components/spatialdisplay/SpatialDisplay.vue
+++ b/src/components/spatialdisplay/SpatialDisplay.vue
@@ -35,14 +35,7 @@
 </template>
 
 <script setup lang="ts">
-import {
-  computed,
-  defineAsyncComponent,
-  ref,
-  useTemplateRef,
-  watch,
-  watchEffect,
-} from 'vue'
+import { computed, defineAsyncComponent, ref, useTemplateRef, watch } from 'vue'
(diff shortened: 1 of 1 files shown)


src/components/spatialdisplay/SpatialDisplay.vue
```diff
@@ -239,6 +239,7 @@ function openLocationsTimeSeriesDisplay(locationIds: string[]) {
     .replace('WithCoordinates', '')
   currentLatitude.value = undefined
   currentLongitude.value = undefined
+  currentLocationIds.value = locationIds
```
wkramer, line 242: This causes two getActions to be fired, and duplicatie timeseries requests.

wkramer (changes requested):

src/components/spatialdisplay/SpatialDisplay.vue
```diff
@@ -204,13 +197,29 @@ const currentLongitude = ref<string>()
...
 
-watchEffect(() => {
-  currentLocationIds.value = props.locationIds?.split(',')
-})
-watchEffect(() => {
-  currentLatitude.value = props.latitude
-  currentLongitude.value = props.longitude
-})
+watch(
+  () => props.locationIds?.split(','),
```
wkramer, line 201: We should watch the query string directly, this would not trigger if the string stays the same.
Do the spil only when assigning the `currentLocationIds.value`.

The `JSON.stringify` is then no longer used.

wkramer (changes requested):

src/components/spatialdisplay/SpatialDisplay.vue
hvangeffen: This did not work since we are checking if the `props.locationIds` are the same as the `currentLocationIds` not to the `props.locationIds` themselves.

wkramer: Currently it does not function correctly when switching topolog nodes or on page refresh.

hvangeffen: Fixed this but it now does not keep the location when switching nodes. Have to rethink this.

PR #1200, opened 2025-02-10
Hide scrollbars on fullscreen dashboard

hvangeffen:
### Description
Hide scrolling for dashboard display which is h100 w100.

### Checklist
    - `rel: new feature`
    - `rel: improvement`
    - `rel: fixes`
    - `rel: ignore`
- [x] Update documentation.
- [x] Update tests.

### DIFF
--- a/src/components/general/DashboardDisplay.vue
+++ b/src/components/general/DashboardDisplay.vue
@@ -121,6 +121,6 @@ watch(
   flex-direction: column;
   width: 100%;
   height: 100%;
-  overflow: auto;
+  overflow: hidden;
 }
 </style>


src/components/general/DashboardDisplay.vue
```diff
@@ -121,6 +121,6 @@ watch(
   flex-direction: column;
   width: 100%;
   height: 100%;
-  overflow: auto;
+  overflow: hidden;
```
wkramer, line 124: Does this not break dashboards that are larger than 100% screen height? I am not seeing any scroll bars with `overflow: auto`.

hvangeffen: We currently do not support dashboards with a height greater than 100%. We need to think about the timeslider placement / scrolling and how to configure it in the first place. The overflow seems to happens on smaller screens.

PR #1213, opened 2025-02-19
Do not round off values in chart mouseover tooltips

ceesvoesenek:
# Description
Previously, values in chart mouseover tooltips were rounded off based on the extent of the axes. In this PR, we make sure that this round-off does not happen, and we use values as they come (appropriately rounded) from the backend.

# Screenshots
Before:
image

After:
image

# Checklist


ceesvoesenek:
Build is currently failing because this requires a new release of `fews-web-oc-charts`.

PR #1230, opened 2025-03-04
Feature/language translation

ElmarGeersHy:
### Description

* added localization functionality via i18n
* added localization for some words in english and german
* default language is en_EN
* added function in configManager for non mandatory environment variables (app-config.json)


TobiasGeeGee:
> @ElmarGeersHy Any thoughts about nesting in the message files? FEWS has `component`.`keyword = value` syntax.

We are also nesting in our other projects, so let's do it.

PR #1235, opened 2025-03-11
Fix wms sometimes not creating a source

hvangeffen:
### Description

This happened when the locations layer was showing and a location is selected and you are trying to go to a different topologynode. In that case the entire wms layer would not be showing. This happened because the styleloaded check considers all layers of the map.

To test:
Select a location with location layer showing and click on another node that includes that location.
Before the wms layer would not show.

### Checklist
    - `rel: new feature`
    - `rel: improvement`
    - `rel: fixes`
    - `rel: ignore`
- [x] Update documentation.
- [x] Update tests.

### DIFF
--- a/src/components/wms/AnimatedRasterLayer.vue
+++ b/src/components/wms/AnimatedRasterLayer.vue
@@ -178,7 +178,7 @@ function getImageSourceOptions(): any {
 }
 
 function createSource() {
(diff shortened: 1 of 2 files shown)


amarkensteijn (changes requested):
The proposed test is still going wrong for me

amarkensteijn (approved):

PR #1244, opened 2025-03-17
Move available workflows fetch to store initialization

hvangeffen:
### Description
When going to a auth redirect the workflows where already being requested. This caused the user to be null and the fetch to not go through. We now only get the workflows the moment the store is initialized. This also ensures we don't fetch if it is not necessary.

### Checklist
    - `rel: new feature`
    - `rel: improvement`
    - `rel: fixes`
    - `rel: ignore`
- [x] Update documentation.
- [x] Update tests.

### DIFF
--- a/src/stores/availableWorkflows.ts
+++ b/src/stores/availableWorkflows.ts
@@ -50,6 +50,9 @@ export const useAvailableWorkflowsStore = defineStore(
       preferredWorkflowIds.value = []
     }
 
+    // Fetch metadata for all available workflows.
+    fetch().catch(() => console.error('Failed to fetch available workflows.'))
+
     return {
       workflows,
       preferredWorkflowIds,
--- a/src/stores/availableWorkflows.ts
+++ b/src/stores/availableWorkflows.ts
@@ -1,54 +1,65 @@
+import { ref, computed } from 'vue'
 import uniq from 'lodash-es/uniq'
 import { defineStore } from 'pinia'
-
 import {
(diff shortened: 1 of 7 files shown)


src/stores/availableWorkflows.ts
```diff
@@ -1,54 +1,67 @@
...
+      preferredWorkflowIds.value = uniqueWorkflowIds
+    }
+
+    function clearPreferredWorkflowIds(): void {
+      preferredWorkflowIds.value = []
+    }
+
+    fetch()
+
+    return {
```
ceesvoesenek, line 55: Could you move this refactor to a separate commit, so the change that actually does something is more clear?

src/stores/availableWorkflows.ts
```diff
@@ -1,54 +1,67 @@
...
-    },
+      uniqueWorkflowIds.forEach((workflowId) => byId(workflowId))
+      preferredWorkflowIds.value = uniqueWorkflowIds
+    }
+
+    function clearPreferredWorkflowIds(): void {
+      preferredWorkflowIds.value = []
+    }
+
+    fetch()
```
ceesvoesenek, line 53: This will silently fail when an error is thrown, maybe at a `.catch`?

ceesvoesenek (commented):
Could you also check whether it is better to revert the previous PR on this problem? Maybe the changes made were an improvement anyway?

ceesvoesenek (approved):

PR #1248, opened 2025-03-19
Add navigation for dashboard and use navigate for ssd's

hvangeffen:
### Description
This fixes navigation in dashboards and switchPanel for ssd's in a topology.

### Checklist
    - `rel: new feature`
    - `rel: improvement`
    - `rel: fixes`
    - `rel: ignore`
- [x] Update documentation.
- [x] Update tests.


wkramer:
Not using child routes brakes the highlighting the selected parent route item (eg. map, report, charts ) when a former child routes was selected. How can we fix this? 
Is there a way to keep using child routes, that is compatible with the use of components in the dashboard?

PR #1265, opened 2025-03-31
Fix bugs that appeared after automatially refreshing time series

amarkensteijn:
### Description

Please provide a brief description of the changes made in this pull request.

If there are any visual changes, please attach screenshots here.

### Link to documentation (if applicable)

Add link to public Github pages (only documentation about configuration)

### Checklist
- [ ] Make the title short and concise
- [ ] Select the correct label to include it in the release notes:
    - `rel: new feature`
    - `rel: improvement`
    - `rel: fixes`
    - `rel: ignore`
- [ ] Update documentation.
- [ ] Update tests.

### DIFF
--- a/src/components/charts/TimeSeriesChart.vue
+++ b/src/components/charts/TimeSeriesChart.vue
@@ -413,7 +413,7 @@ watch(
     }
   },
 )
-watch(props.config, onValueChange)
+watch(() => props.config, onValueChange)
 watch(
   () => props.isLoading,
   (newValue, oldValue) => {


src/components/timeseries/TimeSeriesComponent.vue
```diff
@@ -264,6 +263,11 @@ watch(
 )
 
 watch(isEditing, () => {
+  if (isEditing.value) {
+    useTimeSeriesInterval.pause()
+  } else {
+    useTimeSeriesInterval.resume()
+  }
```
hvangeffen, line 270: Should this also refresh the timeseries on edit cancel?

amarkensteijn: I think so. Essentially, before editing, the timeseries were refreshing continuously. That was stopped when the editting started. Assuming people have the editing mode open for longer than 30 seconds (which is the current interval time), immediatly fetching timeseries again is closest we can get to keeping the automatic refresh going. And even if they cancel before 30 seconds, the worst effect is one fetch before it would otherwise have taken place.

hvangeffen (approved):

PR #1266, opened 2025-03-31
Do not hardcode queryparameters in the time series edit POST

amarkensteijn:
### Description

The 'editUrl' is used directly to POST the edited time series values. The version (required by the TimeSeriesResponseSchema) is taken from the original time series request, instead of performing an additional (only header) timeseries request to get these when editing data.

Currently, this is done by simply adding the version and the timezone through function parameters. It would be cleaner to change the type of the edited data from `Record<string, TimeSeriesEvent[]>`, to `Record<string, TimeSeriesResponse>`, as the time series edit POST expects the body to be of type `TimeSeriesResponse`. Created #1267 to address this.

### Checklist
    - `rel: new feature`
    - `rel: improvement`
    - `rel: fixes`
    - `rel: ignore`
- [ ] Update documentation.
- [ ] Update tests.

### DIFF
--- a/src/components/timeseries/TimeSeriesComponent.vue
+++ b/src/components/timeseries/TimeSeriesComponent.vue
@@ -210,7 +210,14 @@ function isLoading(subplot: ChartConfig, loadingSeriesIds: string[]) {
 }
 
 async function onDataChange(newData: Record<string, TimeSeriesEvent[]>) {
-  await postTimeSeriesEdit(baseUrl, props.config.requests, newData)
+  const seriesHeader = series.value[props.config.requests[0].key].header
+  await postTimeSeriesEdit(
+    baseUrl,
+    props.config.requests,
+    newData,
+    seriesHeader.version ?? '',
+    seriesHeader.timeZone ?? '',
+  )
   lastUpdated.value = new Date()
 }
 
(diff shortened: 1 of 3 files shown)


src/components/timeseries/TimeSeriesComponent.vue
```diff
@@ -211,7 +211,14 @@ function isLoading(subplot: ChartConfig, loadingSeriesIds: string[]) {
...
 
 async function onDataChange(newData: Record<string, TimeSeriesEvent[]>) {
-  await postTimeSeriesEdit(baseUrl, props.config.requests, newData)
+  const seriesHeader = series.value[props.config.requests[0].key].header
+  await postTimeSeriesEdit(
+    baseUrl,
+    props.config.requests,
+    newData,
+    seriesHeader.version ?? '',
+    seriesHeader.timeZone ?? '',
```
hvangeffen, line 220: Do we want empty string here? This parameter is optional I think.

amarkensteijn: Yes. This is one of the things that should be fixed in #1267. Ideally, we would mode the version and timezone one level up, to the part that creates the newData. And that is exactly what #1267 is about.

PR #1277, opened 2025-04-04
Show child locations only when zooming in

hvangeffen:
### Description
Adds the functionality that parents are always visible and children are occluded by other locations. 

### Checklist
    - `rel: new feature`
    - `rel: improvement`
    - `rel: fixes`
    - `rel: ignore`
- [x] Update documentation.
- [x] Update tests.

### DIFF
--- a/src/components/map/MapComponent.vue
+++ b/src/components/map/MapComponent.vue
@@ -13,6 +13,7 @@
     :touchPitch="false"
     :attributionControl="false"
     :bounds="bounds"
+    :fadeDuration="0"
   >
     <mgl-attribution-control position="top-right" :compact="true" />
     <mgl-scale-control position="bottom-right" />
--- a/src/components/wms/LocationsLayer.vue
+++ b/src/components/wms/LocationsLayer.vue
@@ -1,53 +1,36 @@
 <template>
   <mgl-geo-json-source :source-id="locationsSourceId" :data="geojson">
-    <mgl-fill-layer
-      :layer-id="locationsFillLayerId"
-      :paint="paintFillSpecification"
-      :filter="['==', '$type', 'Polygon']"
+    <LocationsFillLayer
+      :layerId="locationsFillLayerId"
+      :selectedLocationIds="selectedLocationIds"
+      :isDark="isDark"
+      :hoveredStateId="hoveredStateId"
     />
-    <mgl-symbol-layer
-      :layer-id="locationsSymbolLayerId"
-      :layout="layoutSymbolSpecification"
-      :paint="paintSymbolSpecification"
-      :filter="['all', ['has', 'iconName'], ['==', '$type', 'Point']]"
-    />
-    <mgl-circle-layer
-      :layer-id="locationsCircleLayerId"
-      :paint="paintCircleSpecification"
-      :filter="['all', ['!has', 'iconName'], ['==', '$type', 'Point']]"
-    />
-    <mgl-symbol-layer
+    <LocationsSymbolLayer :layerId="locationsSymbolLayerId" />
+    <LocationsCircleLayer :layerId="locationsCircleLayerId" />
+    <LocationsTextLayer
       v-if="showNames"
-      :layer-id="locationsTextLayerId"
-      :layout="layoutTextSpecification"
-      :paint="paintTextSpecification"
-      :filter="['==', '$type', 'Point']"
+      :layerId="locationsTextLayerId"
+      :isDark="isDark"
     />
   </mgl-geo-json-source>
-  <mgl-marker
-    v-for="coordinates in selectedLocationsCoordinates"
-    :coordinates="coordinates"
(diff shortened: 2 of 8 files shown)


src/components/wms/LocationsLayer.vue
```diff
@@ -6,13 +6,19 @@
       :isDark="isDark"
       :hoveredStateId="hoveredStateId"
     />
-    <LocationsSymbolLayer :layerId="locationIds.layer.symbol" />
+
     <LocationsCircleLayer :layerId="locationIds.layer.circle" />
-    <LocationsTextLayer
-      v-if="showNames"
```
amarkensteijn, line 20: By removing the `v-if`, the LocationsTextLayer will always exist, even if the text will not be shown because the `locationName` of the `feature` will be an empty string. Was this an intended change? Are there downsides to always having this layer?

src/components/wms/locations/LocationsTextLayer.vue
```diff
@@ -28,9 +33,7 @@ const layout = {
   'text-justify': 'auto',
   'text-variable-anchor': ['right', 'left'],
   'text-max-width': 15,
-  // When overlap is false sort order has to be inverted for some reason:
```
amarkensteijn: This seems like a relevant comment. Even though the `invertedSortKey` is nice, it is still weird that we need to do that, so maybe keep this comment somewhere

src/components/wms/locations/LocationsSymbolLayer.vue
```diff
@@ -1,21 +1,64 @@
...
   'icon-image': ['get', 'iconName'],
   'symbol-sort-key': ['get', 'sortKey'],
 }
 
-const filter = ['all', ['has', 'iconName'], ['==', '$type', 'Point']]
+const childLayout = {
+  ...baseLayout,
+  'icon-overlap': 'never',
+  'text-optional': true,
+  'text-field': ['get', 'locationName'],
```
amarkensteijn, line 38: Wait... so.... The `locationName` of a parent is part of the `LocationsTextLayer`, but the `locationName` of a child is part of the `LocationsSymbolLayer`? That is counterintuitive; I'd expect all labels to be part of the text layer.

If this is required in order to have the zooming function as we want, then we should at least make it clear somewhere/somehow that the text layer only contains the parent text. Could be as simple as renaming that one from `LocationsTextLayer` to `ParentLocationsTextLayer` or `LocationsParentTextLayer`.

amarkensteijn (changes requested):
image
These seem to be childs (at least one of them), as the top right pump (?) sometimes is not visible. I'd expect it not to be visible at this zoom level, since the icons overlap. Is this a bug, or is this weird map-libre behaviour?

src/components/map/MapComponent.vue
```diff
@@ -13,6 +13,7 @@
     :touchPitch="false"
     :attributionControl="false"
     :bounds="bounds"
+    :fadeDuration="0"
```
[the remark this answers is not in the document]
hvangeffen: Increased it to 100ms to fix the above issue

src/lib/map/locations.ts
```diff
@@ -1,14 +1,25 @@
 import { getLayerId, getSourceId } from '@/lib/map/utils'
 
-export const locationsCircleLayerId = getLayerId('location-circle')
-export const locationsSymbolLayerId = getLayerId('location-symbol')
-export const locationsTextLayerId = getLayerId('location-text')
-export const locationsFillLayerId = getLayerId('location-fill')
-export const locationsSourceId = getSourceId('location')
+export const locationIds = {
```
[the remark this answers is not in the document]
hvangeffen: Renamed to locationMapIds

src/lib/map/locations.ts
```diff
@@ -1,14 +1,25 @@
...
-  locationsSymbolLayerId,
-  locationsTextLayerId,
-  locationsFillLayerId,
+  locationIds.layer.circle,
+  locationIds.layer.symbol,
+  locationIds.layer.text,
+  locationIds.layer.fill,
+]
+
+// NOTE: When multiple layers are clicked the order of the layers here is important.
```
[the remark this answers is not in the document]
hvangeffen: Made it object.values to make it different from below

src/components/wms/LocationsLayer.vue
hvangeffen: Important for ordering and not a problem in maplibre.

src/components/wms/locations/LocationsTextLayer.vue
hvangeffen: Added back comments

src/components/wms/LocationsLayer.vue
amarkensteijn: Good to know

amarkensteijn (approved):

PR #1279, opened 2025-04-04
Fix startPanel being overwritten and tooltip not loading when no timeseries is enabled

hvangeffen:
### Checklist
    - `rel: new feature`
    - `rel: improvement`
    - `rel: fixes`
    - `rel: ignore`
- [x] Update documentation.
- [x] Update tests.


amarkensteijn (changes requested):
If I refresh, let's say /topology/node/all_systems/map/location/211, it still opens on the timeseries table, and not on the meta data panel

hvangeffen:
> If I refresh, let's say /topology/node/all_systems/map/location/211, it still opens on the timeseries table, and not on the meta data panel

Fixed, issue was that the componentSettings are loaded in async so it should have a watch

amarkensteijn (approved):

PR #1294, opened 2025-04-10
Add dashboard interaction

hvangeffen:
### Description

Adds support for ssd web oc dashboard actions and add interaction between map and charts in the same group.

Removes currentLocationIds and current lat lng and makes it work with just the state of the props.locationIds.
This is done by providing them from the `from` route in displayTabs to.
Now the chart panel is closed when the locationIds are not a subset of the newly fetched locations instead of "opened" in the current inverted situation.

### Checklist
    - `rel: new feature`
    - `rel: improvement`
    - `rel: fixes`
    - `rel: ignore`
- [x] Update documentation.
- [x] Update tests.


src/lib/topology/displayTabs.ts
```diff
@@ -55,10 +56,22 @@ export function displayTabsForNode(
...
+        if (from?.params.locationIds) {
+          tab.to.name = 'TopologySpatialTimeSeriesDisplay'
+          tab.to.params.locationIds = from.params.locationIds
+        }
+
+        if (from?.params.latitude && from?.params.longitude) {
+          tab.to.name = 'TopologySpatialTimeSeriesDisplayWithCoordinates'
+          tab.to.params.latitude = from.params.latitude
+          tab.to.params.longitude = from.params.longitude
+        }
```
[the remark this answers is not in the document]
hvangeffen: Done

src/components/general/DashboardDisplay.vue
```diff
@@ -75,6 +75,17 @@ function setupDates() {
...
 }
 
+const actionId = ref<string>()
+function onDashboardAction(action: SsdActionResult) {
+  actionId.value = action.actionId
+}
+
+function getDashboardItem(items: WebOCDashboardItem[]) {
+  return (
+    items.find((item) => item.actionIds?.includes(actionId.value)) ?? items[0]
```
amarkensteijn: As discussed before, it does not make sens to me to do this using `actionIds`. I'd expect something like `targetIds` or `targetElement` instead. See also [this DWO-1016 comment](https://issuetracker.deltares.nl/browse/DWO-1016?focusedId=986631&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-986631).

src/components/ssd/SchematicStatusDisplay.vue
```diff
@@ -129,31 +129,28 @@ const mobile = computed(() => {
 
 function onAction(event: CustomEvent<SsdActionEventPayload>): void {
   const { panelId, objectId, results } = event.detail
-  const now = new Date()
-  if (results.length === 0) {
-    alertsStore.addAlert({
-      id: `undefined-action-${now.toISOString()}`,
-      type: 'error',
-      message: 'No left click actions defined for this object',
-    })
```
amarkensteijn, line 138: Is there an option of the frontend/cursor indicating that we can click on something, but then it not having any action at all? If so, I think this error dialog is good to have.

src/components/ssd/SchematicStatusDisplay.vue
```diff
@@ -133,12 +139,13 @@ function onAction(event: CustomEvent<SsdActionEventPayload>): void {
...
   const result = results[0]
+  const request = result.requests?.[0]
   switch (result.type) {
     case 'PDF':
-      window.open(new URL(result.requests[0].request))
+      if (request) window.open(new URL(request.request))
       break
     case 'SSD':
-      switchPanel(result.requests[0])
+      if (request) switchPanel(request)
```
amarkensteijn: Do we also want one of those dialogs if `request` is undefined?

src/components/general/DashboardItem.vue
```diff
@@ -51,7 +51,7 @@ const topologyNodesStore = useTopologyNodesStore()
 
 const componentItem = computed(() => {
   const hasActionId =
-    props.actionId && props.item.actionIds?.includes(props.actionId)
+    !props.actionId || props.item.actionIds?.includes(props.actionId)
```
amarkensteijn, line 54: `const hasActionId = !props.actionId` ?!?! Please rename the `hasActionId` to represent what it actually is.

src/components/general/DashboardDisplay.vue
hvangeffen: For now this is how the backend is so we can pick this up later in DWO-1016

src/components/ssd/SchematicStatusDisplay.vue
hvangeffen: We now have two requests that resolve in different orders. For either one we don't know whether the other is also empty.

src/components/ssd/SchematicStatusDisplay.vue
hvangeffen: Do we?

src/components/general/DashboardItem.vue
hvangeffen: Renamed

src/components/ssd/SchematicStatusDisplay.vue
amarkensteijn: Don't know... It is just that we do give an alert if we cannot handle that type of action yet. But not having a request might be a different use case altogether

src/components/ssd/SchematicStatusDisplay.vue
amarkensteijn: Aha. That complicates things

PR #1341, opened 2025-04-30
Route to alert station on alert click

cees-wolfs-vt:
See Issue

### DIFF
--- a/src/components/thresholds/ThresholdSummary.vue
+++ b/src/components/thresholds/ThresholdSummary.vue
@@ -5,18 +5,18 @@
     flat
     density="compact"
     :ripple="false"
-    @click="onExpansionPanelToggle"
+    @click="onPanelClick"
     @mouseenter="onMouseEnter"
     @mouseleave="onMouseLeave"
   >
     <template v-for="(crossing, index) in crossings">
       <ThresholdsParameter
-        v-if="index === 0 || expanded"
+        v-if="index === 0 || isSelected"
         :key="crossing.parameterId"
         :crossing="crossing"
         :time-is-relative="relativeFormat"
       >
-        <template #append v-if="crossings.length > 1 && !expanded">
+        <template #append v-if="crossings.length > 1 && !isSelected">
           <v-chip
             size="small"
             density="compact"
@@ -31,9 +31,9 @@
 </template>
 <script setup lang="ts">
 import ThresholdsParameter from '@/components/thresholds/ThresholdsField.vue'
+import { NavigateRoute } from '@/lib/router/types'
 import { LevelThresholdCrossings } from '@deltares/fews-pi-requests'
 import { ref } from 'vue'
-import { useRouter } from 'vue-router'
 
 interface Props {
   crossings: LevelThresholdCrossings[]
@@ -42,12 +42,11 @@ interface Props {
 
 const props = defineProps<Props>()
 const relativeFormat = ref(false)
-const router = useRouter()
 
-const expanded = defineModel<boolean>('expanded', {
-  required: false,
-  default: false,
-})
+interface Emits {
+  navigate: [to: NavigateRoute]
+}
+const emit = defineEmits<Emits>()
 
 function onMouseEnter() {
   relativeFormat.value = true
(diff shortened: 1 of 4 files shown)


src/components/thresholds/ThresholdSummary.vue
```diff
@@ -1,6 +1,7 @@
 <template>
   <v-card
     :border="true"
+    :color="isSelected ? 'rgba(0, 0, 0, 0.13)' : 'transparent'"
```
hvangeffen, line 4: This is not really visible in dark mode

src/components/thresholds/ThresholdSummary.vue
```diff
@@ -58,5 +62,12 @@ function onExpansionPanelToggle() {
   if (window.getSelection()?.toString() === '') {
     expanded.value = !expanded.value
   }
+  // Route to open the crossing in the map
+  router.push({
+    name: 'TopologySpatialTimeSeriesDisplay',
+    params: {
+      locationIds: props.crossings[0].locationId,
+    },
+  })
```
hvangeffen, line 71: This won't work inside dashboards, we now do routing by sending a navigate event upwards. In the end I think we would like to receive the event in the `TopologyDisplayView`, something like:
```
    <ThresholdsControl :topologyNode="topologyNode" @navigate="onNavigate" />
```

Example in `SpatialDisplay`

src/components/thresholds/ThresholdsPanel.vue
```diff
@@ -68,8 +71,23 @@ interface Props {
...
 const expandedItems = ref<Record<string, boolean>>({})
+const route = useRoute()
+const virtualScroll = ref<any>(null)
+
+function getRouteLocationIds(): string[] {
+  if (typeof route.params.locationIds === 'string') {
+    return route.params.locationIds.split(',')
+  }
+  return Array.isArray(route.params.locationIds) ? route.params.locationIds : []
+}
```
hvangeffen, line 83: These locationIds should also be passed as props from `TopologyDisplayView` onwards

src/components/thresholds/ThresholdsPanel.vue
```diff
@@ -68,8 +71,23 @@ interface Props {
...
+
+const isLocationSelected = computed(() => {
+  const routeLocationIds = getRouteLocationIds()
+  return (locationId: string): boolean => {
+    return routeLocationIds.includes(locationId)
+  }
+})
```
hvangeffen, line 90: I think this can just be a function

src/components/thresholds/ThresholdsPanel.vue
```diff
@@ -68,8 +71,23 @@ interface Props {
 const props = defineProps<Props>()
 
 const selectedWarningLevelIds = defineModel<string[]>('selectedWarningLevelIds')
-
 const expandedItems = ref<Record<string, boolean>>({})
+const route = useRoute()
+const virtualScroll = ref<any>(null)
```
hvangeffen, line 76: This can be `useTemplateRef('virtualScroll')`

src/components/thresholds/ThresholdsPanel.vue
```diff
@@ -82,6 +100,30 @@ const groupedCrossings = computed(() => {
...
+}
+
+watch([() => route.params.locationIds, groupedCrossings], () => {
+  scrollToSelectedItem()
+})
+
+// Scroll to the selected item when we directly navigate to the page
+onMounted(() => {
+  nextTick(() => scrollToSelectedItem())
+})
```
hvangeffen, line 126: Do we need nextTick here since the function already has an await nextTick?

src/components/thresholds/ThresholdsPanel.vue
```diff
@@ -46,6 +47,7 @@
           <ThresholdSummary
             v-model:expanded="expandedItems[crossingsGroup[0].locationId]"
```
hvangeffen, line 48: Maybe we should remove expanded and only have a single item open based on the locationIds.
It is now quite weird when you want to expand and un-expand an item.

hvangeffen (approved):

PR #1352, opened 2025-05-07
Use time series thinning for charts but not for tables

ceesvoesenek:
### Description
The FEWS PI Webservice supports "thinning" long time series to reduce computational load when plotting them. This is useful for charts, but makes the data irregularly spaced which is unsuitable for display in a table.

Currently, data for charts and tables are shared, so we cannot use thinning. This PR separates fetching data for charts and tables, and adds thinning to chart data.

### Checklist

### DIFF
--- a/src/components/charts/TimeSeriesChart.vue
+++ b/src/components/charts/TimeSeriesChart.vue
@@ -73,6 +73,11 @@ const props = withDefaults(defineProps<Props>(), {
   },
 })
 
+interface Emits {
+  'update:x-domain': [old: [Date, Date], new: [Date, Date]]
+}
+const emit = defineEmits<Emits>()
+
 let thresholdLines!: ThresholdLine[]
 let thresholdLinesVisitor!: AlertLines
 let axis!: CartesianAxes
@@ -96,6 +101,13 @@ onMounted(() => {
       props.verticalProfile ? 1200 : null,
       axisOptions,
     )
+    axis.addEventListener('update:x-domain', (event) => {
+      emit(
+        'update:x-domain',
+        event.old as [Date, Date],
+        event.new as [Date, Date],
+      )
+    })
 
     // Keep margin in sync with axis.margin
     axis.margin = new Proxy(axis.margin, {
--- a/src/components/charts/TimeSeriesChart.vue
+++ b/src/components/charts/TimeSeriesChart.vue
@@ -14,7 +14,15 @@
 </template>
 
 <script setup lang="ts">
-import { ref, watch, onMounted, onBeforeUnmount, nextTick, computed } from 'vue'
+import {
+  ref,
+  watch,
+  onMounted,
+  onBeforeUnmount,
+  nextTick,
+  computed,
+  useTemplateRef,
+} from 'vue'
 import {
   AlertLines,
   ChartArea,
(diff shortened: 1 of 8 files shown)


src/components/timeseries/TimeSeriesComponent.vue
```diff
@@ -200,13 +227,16 @@ function isLoading(subplot: ChartConfig, loadingSeriesIds: string[]) {
 }
 
 async function onDataChange(newData: Record<string, TimeSeriesEvent[]>) {
-  const seriesHeader = series.value[props.config.requests[0].key].header
+  const seriesKey = props.config.requests[0]?.key
+  const seriesHeader = seriesKey
```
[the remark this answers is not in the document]
ceesvoesenek: Not very useful suggestion, ignoring.

src/components/timeseries/TimeSeriesComponent.vue
```diff
@@ -298,6 +326,37 @@ function onConfirmationLeave() {
...
+  // Do not refetch if we are zoomed in past the point where thinning has
+  // any effect.
+  const isDomainLargeEnough =
+    getRange(newDomain) >= MIN_HOURS_FOR_REFETCH * 60 * 60 * 1000
+  if (!isDomainLargeEnough) return
+
+  const zoomRatio = getRange(newDomain) / getRange(oldDomain)
+
+  // When zooming out, always refresh the data; we need more data than we had
+  // before.
```
hvangeffen, line 346: If zoom out after we don't zoom in enough (so no refetch) we still do a fetch.

src/components/charts/TimeSeriesChart.vue
```diff
@@ -96,6 +101,13 @@ onMounted(() => {
       props.verticalProfile ? 1200 : null,
       axisOptions,
     )
+    axis.addEventListener('update:x-domain', (event) => {
+      emit(
+        'update:x-domain',
+        event.old as [Date, Date],
+        event.new as [Date, Date],
+      )
+    })
```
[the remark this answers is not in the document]
ceesvoesenek: That would indeed be very nice and would solve a lot of typing annoyances...

src/components/timeseries/TimeSeriesComponent.vue
ceesvoesenek: This is fixed by comparing the new domain with the domain of the currently fetched data, instead of the previous axis domain.

src/components/charts/TimeSeriesChart.vue
```diff
@@ -306,30 +322,46 @@ const refreshChart = () => {
...
+    // Update each matching chart
+    charts.forEach((chart) => {
+      // If any chart has previous data, we should not rescale the axes.
+      needsAxisRescale &&= chart.data === undefined || chart.data.length === 0
+      chart.data = data
+    })
   })
+  if (needsAxisRescale) {
+    // Autoscale only the y-axis, the x-axis should keep the domain set from
+    // the start and end time or zooming.
```
hvangeffen, line 341: I now get that when I zoom in I have to double click twice to get back to my original zoom:
### Original
image
### Zoom
image
### After first doubleclick
image

After second doubleclick I go back to original

ceesvoesenek: Should be fixed

hvangeffen (approved):

PR #1359, opened 2025-05-14
Allow panning in time series charts

ceesvoesenek:
### Description
Holding the shift button and dragging with the left mouse button now pans the time axis and vertical of charts. If multiple charts are configured for the same location, the time axis of all charts will be panned synchronously, but the y-axis of only the initially clicked chart is panned. Upon finishing panning, the time series data are updated to span the new time range.

### Checklist
- [x] Select the correct label to include it in the release notes.

### DIFF
--- a/src/components/charts/TimeSeriesChart.vue
+++ b/src/components/charts/TimeSeriesChart.vue
@@ -57,6 +57,7 @@ import { difference, uniq } from 'lodash-es'
 import { getMatchingIndexedString, type Tag } from '@/lib/charts/tags'
 import { type ChartsSettings } from '@/lib/topology/componentSettings'
 import { getAxisOptions } from '@/lib/charts/axisOptions'
+import { PanHandler } from '@deltares/fews-web-oc-charts'
 
 interface Props {
   config?: ChartConfig
@@ -64,6 +65,7 @@ interface Props {
   highlightTime?: Date
   isLoading?: boolean
   zoomHandler?: ZoomHandler
+  panHandler?: PanHandler
   verticalProfile?: boolean
   forecastLegend?: string
   settings: ChartsSettings['timeSeriesChart']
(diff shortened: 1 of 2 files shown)


src/components/charts/TimeSeriesChart.vue
```diff
@@ -129,6 +134,12 @@ onMounted(() => {
       ? new VerticalMouseOver(undefined, (value: number) => value.toString())
       : new MouseOver(undefined, (value: number) => value.toString())
     const zoom = props.zoomHandler ?? new ZoomHandler(WheelMode.NONE)
+    const panX = props.panHandler ?? new PanHandler()
```
hvangeffen, line 134: What does the default pan handler do?

src/components/timeseries/TimeSeriesComponent.vue
```diff
@@ -333,18 +347,21 @@ function shouldRefetchAfterDomainUpdate(newDomain: [Date, Date]): boolean {
...
 
   const newDomainRange = newDomain[1].getTime() - newDomain[0].getTime()
+  const oldDomainRange =
+    chartOptions.value.endTime.getTime() -
+    chartOptions.value.startTime.getTime()
+  const zoomRatio = newDomainRange / oldDomainRange
+
+  // Detect panning; we should always refetch if we are panning.
+  const isPanning = Math.abs(zoomRatio - 1.0) < 1e-6
+  if (isPanning) return true
```
hvangeffen, line 357: Can this ever happen when we get new data or a different dataset with the same domain?

ceesvoesenek: If we get new data, we are refetching anyway; in that case the refetch is not coming from an updated domain

src/components/charts/TimeSeriesChart.vue
ceesvoesenek: Good point, the default should probably be no pan handler at all

src/components/timeseries/TimeSeriesComponent.vue
hvangeffen: No double fetch, new data causing a new domain?

ceesvoesenek: Should not be the case because if the first fetch set the domain and it did not change, no event is fired in `fews-web-oc-charts`

hvangeffen (approved):

PR #1434, opened 2025-06-11
When the user login expired try a silent sign in on app init or reload

wkramer:
### Description

When the user login expired try a silent sign in on app init or reload.


hvangeffen (approved):
Nice, I was thinking of something like this before. Right now this requires a page refresh, do we maybe want to check this each time `transformRequestAuth` is called?

PR #1445, opened 2025-06-12
Allow scroll wheel zooming in charts while holding Shift

ceesvoesenek:
### Description
This PR adds the possibility to use the scroll wheel to zoom in and out while holding the Shift-key (which also enters panning mode). Which axes (if any) are zoomed with the scroll wheel can be configured from the user settings store.

### Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "@deltares/fews-pi-requests": "^1.19.3",
     "@deltares/fews-ssd-requests": "^1.1.1",
     "@deltares/fews-ssd-webcomponent": "^1.1.1",
-    "@deltares/fews-web-oc-charts": "^4.0.0-alpha.11",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -21,7 +21,7 @@
     "@deltares/fews-pi-requests": "^1.17.0",
     "@deltares/fews-ssd-requests": "^1.1.1",
     "@deltares/fews-ssd-webcomponent": "^1.1.1",
-    "@deltares/fews-web-oc-charts": "^4.0.0-alpha.8",
+    "@deltares/fews-web-oc-charts": "^4.0.0-alpha.9",
```
hvangeffen, line 24: Something seems to have broken in alpha.9 release as the share zooming and resetting does not sync up well anymore.

hvangeffen (changes requested):

hvangeffen (approved):

PR #1537, opened 2025-07-15
Fix autoselection of first topology Node and add tests for this

cees-wolfs-vt:


src/views/TopologyDisplayView.vue
```diff
@@ -494,7 +494,9 @@ onBeforeRouteUpdate(reroute)
 
 function reroute(to: RouteLocationNormalized, from?: RouteLocationNormalized) {
   if (!to.params.nodeId) {
-    const firstSubNodeId = subNodes.value[0].id
+    const firstSubNodeId = topologyNodesStore.getFirstLeafNodeForId(
+      subNodes.value[0].id,
+    )?.id
     if (firstSubNodeId) {
       to.params.nodeId = firstSubNodeId
       return to
```
hvangeffen, line 502: Maybe make this `return reroute(to)` instead of the change above

hvangeffen (approved):

PR #1588, opened 2025-08-28
Fix bug in navigating between topologies with and without leafnodebuttons

cees-wolfs-vt:
Fix bug in navigating between topologies with and without leafnodebuttons DWO-1170

### Checklist
    - `rel: new feature`
    - `rel: improvement`
    - `rel: fixes`
    - `rel: ignore`
- [ ] Update documentation.
- [ ] Update tests.

### DIFF
--- a/src/components/general/LeafNodeButtons.vue
+++ b/src/components/general/LeafNodeButtons.vue
@@ -1,7 +1,7 @@
 <template>
   <v-menu v-if="items.length > 0">
     <template #activator="{ props }">
-      <v-btn :variant v-bind="props" class="text-none">
+      <v-btn :variant v-bind="props" class="text-none" aria-label="Leaf node">
         <v-icon start>mdi-filter-variant</v-icon>
         {{ activeNode?.name || 'All' }}
         <v-icon v-if="items.length > 1" end>mdi-menu-down</v-icon>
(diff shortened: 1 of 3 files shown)


src/views/TopologyDisplayView.vue
```diff
@@ -336,8 +336,9 @@ watch(
 const topologyComponentConfig = computed(() => {
   const component = props.topologyId
     ? configStore.getComponentById(props.topologyId)
-    : configStore.getComponentByType('TopologyDisplay')
-  return component as WebOcTopologyDisplayConfig | undefined
+    : undefined
+  const config = component as WebOcTopologyDisplayConfig | undefined
+  return config ?? configStore.getComponentByType('TopologyDisplay')
```
hvangeffen, line 341: I updated this to now fallback to getComponentByType in case topologyId is not found. Maybe make a util function of this and reuse it below

hvangeffen (commented):
Looks good, I did notice a seperate issue where when switchin topologyId's we don't go to the first available node such as on page refresh. Might be nice to fix here as well, otherwise we could create a new issue.

cees-wolfs-vt:
> Looks good, I did notice a seperate issue where when switchin topologyId's we don't go to the first available node such as on page refresh. Might be nice to fix here as well, otherwise we could create a new issue.

I cannot reproduce this, maybe we can take a quick look together

PR #1590, opened 2025-08-29
Add scrolling to data table of timeseries

cees-wolfs-vt:
Add scrolling to data table of timeseries using before and after start and end time counts


wkramer:
@hvangeffen When reaching the bottom of the table, the progress indicator for loading more data appears and does not stop.

PR #1596, opened 2025-09-11
Redirect to first topology on load

cees-wolfs-vt:
See [DWO-1174](issuetracker.deltares.nl/browse/DWO-1174)

### DIFF
--- a/src/router/index.ts
+++ b/src/router/index.ts
@@ -317,7 +317,9 @@ async function handleAuthorization(to: RouteLocationNormalized) {
 async function addDynamicRoutes() {
   const store = useConfigStore()
   await store.setFewsConfig()
-  Object.values(store.components).forEach((component: any) => {
+  const configuredComponents = Object.values(store.components)
+
+  configuredComponents.forEach((component: any) => {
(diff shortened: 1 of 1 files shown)


src/views/AboutView.vue
```diff
@@ -81,6 +83,17 @@ onMounted(async () => {
     await webServiceProvider.getVersion()
   ).version
 })
+
+watch(
```
wkramer, line 87: This should not be placed in the `AboutView.vue`, but in `src/router/index.ts`.
This seems to be the place:
```typescript
  if (!routesAreInitialized) {
    await addDynamicRoutes()
    routesAreInitialized = true
    if (redirect) {
      return redirect
    }
    return to
  }
```

wkramer (changes requested):

src/views/AboutView.vue
cees-wolfs-vt: I put it in addDynamicRoutes as I could add it there with minimal edits without breaking the existing defaultComponent logic

wkramer (approved):

PR #1600, opened 2025-09-16
Show select in user settings menu when setting has 4 options or more

wkramer:
### Description

Show select in user settings menu when setting has 4 options or more.
Add icons for scroll wheel zoom mode options.

Screenshot 2025-09-16 155628

### DIFF
--- a/src/components/user-settings/UserSettingsMenu.vue
+++ b/src/components/user-settings/UserSettingsMenu.vue
@@ -10,8 +10,11 @@
         :disabled="setting?.disabled"
       >
         <v-list-item-subtitle>{{ setting.label }} </v-list-item-subtitle>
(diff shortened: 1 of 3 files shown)


src/components/user-settings/UserSettingsMenu.vue
```diff
@@ -33,6 +36,37 @@
...
+            hide-details
+            :aria-label="`${setting.label}`"
+            class="my-2"
+            @click.preventDefault
+            @update:modelValue="onValueChange(setting)"
+          >
+            <template #item="{ props }">
+              <v-list-item
+                v-bind="props"
+                :disabled="props?.disabled === true"
```
hvangeffen, line 59: Doesn't v-bind props already do this, or does this miss the undefined case?

PR #1722, opened 2025-12-11
fix: Incomplete multi-character sanitization

wkramer:
### Description

Fix Incomplete multi-character sanitization


hvangeffen (commented):
How come a single replace pass does not get all the style blocks already? The regex itself has /g so it should match all of them right? When would they be nested?

PR #1830, opened 2026-02-06
feat: update measure tools to support dark theme

### LINKED ISSUE #1829: DWO-1247: Update measure tools map paint style and support dark theme

wkramer:
### Description

The map measuring tools style is update to better contrast in dark mode. Also reduces text halo width to be more inline with Carto style labels. Renames "Map tools" to "Measuring tools". 

If there are any visual changes, please attach screenshots here.

### DIFF
--- a/src/components/map/MapToolsControl.vue
+++ b/src/components/map/MapToolsControl.vue
@@ -49,27 +49,31 @@ const removeControl = () => {
 }
 
 onMounted(() => {
-  watch(
-    () => settings.get('ui.map.measuringTools')?.value ?? false,
-    (enabled) => {
-      if (!map) {
-        console.warn('Map is not available')
-        return
-      }
-
-      if (enabled) {
-        if (!measureControl) {
-          addControl()
-        }
-      } else {
-        if (measureControl) {
-          removeControl()
-        }
-      }
-    },
-    { immediate: true },
-  )
+  const enabled = settings.get('ui.map.measuringTools')?.value ?? false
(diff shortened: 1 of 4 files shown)


src/components/map/MapToolsControl.vue
```diff
@@ -5,14 +5,29 @@ import { onBeforeUnmount, watch, onMounted } from 'vue'
 import { useMap } from '@/services/useMap'
 import { useUserSettingsStore } from '@/stores/userSettings'
 import { MaplibreMeasureControl } from '@watergis/maplibre-gl-terradraw'
+import {
+  getLineLayerLabelSpec,
+  getModeOtions,
+  getPointLayerLabelSpec,
+  getPolygonLayerSpec,
+} from '@/lib/map/terraDraw'
+import { useDark, useStorage } from '@vueuse/core'
```
hvangeffen, line 14: This `useDark` should be replaced with an import from `@/services/useDark` as it looks at the theme choice instead and has better performance on change

src/components/map/MapToolsControl.vue
```diff
@@ -37,27 +71,31 @@ const removeControl = () => {
...
-    { immediate: true },
-  )
-})
+    } else {
+      if (measureControl) {
+        removeControl()
+      }
+    }
+  },
+)
```
hvangeffen, line 98: Would this work with `immediate: true` on the watch instead of the onMounted?

PR #1839, opened 2026-02-06
Fix double click on map bug

cees-wolfs-vt:
Fix double click not opening time series chart

### DIFF
--- a/src/components/spatialdisplay/SpatialDisplay.vue
+++ b/src/components/spatialdisplay/SpatialDisplay.vue
@@ -298,6 +298,9 @@ const hideMap = computed(() => {
 function onLocationsChange(locationIds: string[]): void {
   if (locationIds.length) {
     openLocationsTimeSeriesDisplay(locationIds)
+  } else if (props.longitude && props.latitude) {
+    // Opening a coordinates display so do not close time series display
+    return
   } else {
     closeTimeSeriesDisplay()
   }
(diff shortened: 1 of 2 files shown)


src/components/spatialdisplay/SpatialDisplay.vue
```diff
@@ -371,6 +364,10 @@ function openCoordinatesTimeSeriesDisplay(latitude: number, longitude: number) {
 }
 
 function closeTimeSeriesDisplay(): void {
+  if (openingCoordinatesTimeSeriesDisplay.value) {
+    openingCoordinatesTimeSeriesDisplay.value = false
+    return
+  }
```
hvangeffen: The root problem was that:
```ts
watch(
  () => props.selectedLocationIds,
  (ids) => {
    if (ids.length === 0 && selectedItems.value.length === 0) return
    return (selectedItems.value = ids)
  },
  { immediate: true, deep: true },
)
```
in `LocationsSeachControl.vue` should not have immediate `true`.
This would cause the selectedItems to trigger an empty array changeLocationIds emit on mount, causing this to trigger a close: 
```ts
function onLocationsChange(locationIds: string[]): void {
  if (locationIds.length) {
    openLocationsTimeSeriesDisplay(locationIds)
  } else {
    closeTimeSeriesDisplay()
  }
}
```
Maybe just removing the immediate would be better.

wkramer: Agree, most of the `immediate: true` uses I see in PR's are not required.

PR #1929, opened 2026-04-07
Feat: overlay layer options menu

### LINKED ISSUE #1834: DWO-1261: Layer options menu

wkramer:
### Description

Add option to change the overlay order and opacity.

### Screenshots

Screenshot_15-4-2026_13050_localhost


hvangeffen (dismissed):
I still have to fix the dragging of overlays. Does not seem to update correctly within the maintainlayerorder component. Sometimes the WMS goes above.

wkramer:
@hvangeffen We should add an entry for the wms layer in the overlay panel (which should be renamed to layer options?).

hvangeffen:
> @hvangeffen We should add an entry for the wms layer in the overlay panel (which should be renamed to layer options?).

@wkramer I do still have some questions. We now use a single store and add overlays which have not yet been added to the store to it. This now does not interact in any way with the scoped component settings of a given node, which itself provides an ordered subset of overlays. Do we still want this? And what happens when the backend no longer provides a certain overlay? Currently that overlay would still be stored in the store.

wkramer:
@hvangeffen Maybe we should have a reset button? If we switch from backends we will have non existing overlays in the menu. The menu should only list overlays from the component settings. But it should remember order and opacity of layers that are not available for the current topology node.

PR #1945, opened 2026-04-15
feat: suppression of empty timeseries displays

### LINKED ISSUE #1765: Hydrotec: Suppression of empty time series displays
## Description

If a filter displays locations for which no time series data exist in the database for a given parameter, it would be preferable to not show any display at all if the location is clicked on in the display, instead of an empty time series display (optionally controlled by a setting such as “show empty displays: true/false”).

---

## Use Case / Example

In a WebOC topology node, we want to display gauge locations. There are gauges with water level data and external gauges that should only be linked to. If empty displays are shown for these external gauges, it appears as if there is a data outage.

We could display the gauges in a separate node and hide the time series charts there through the WebOCComponentSettings. However, in an overall overview node that contains both internal and external gauges, the displays for the empty external ones would still be shown. 

---

## Criticality / Deadline

It would be nice to have this feature within the next ~2 months

AntonPetry-Hydrotec:
close #1765

### DIFF
--- a/src/components/timeseries/TimeSeriesWindowComponent.vue
+++ b/src/components/timeseries/TimeSeriesWindowComponent.vue
@@ -83,7 +83,6 @@ import { computed, ref, StyleValue, watch } from 'vue'
 import { useUserSettingsStore } from '@/stores/userSettings'
 import type {
   filterActionsFilter,
-  Location,
   timeSeriesGridActionsFilter,
 } from '@deltares/fews-pi-requests'
 
@@ -95,7 +94,6 @@ interface Props {
   informationContent?: string | null
   filter?: filterActionsFilter | timeSeriesGridActionsFilter
   settings: ChartsSettings
-  location?: Location
 }
 
 const props = defineProps<Props>()
@@ -168,16 +166,6 @@ watch(
   { immediate: true },
 )
 
-function getDisplayEnabled(
-  settingsEnabled: boolean,
-  attributeId: string | undefined,
-): boolean {
-  if (!attributeId) return settingsEnabled
-  const attr = props.location?.attributes?.find((a) => a.id === attributeId)
-  if (attr === undefined) return settingsEnabled
-  return attr.value !== 'false'
-}
-
 const displayTypeItems = computed<DisplayTypeItem[]>(() => {
   const noElevationCharts = !(
     (props.elevationChartDisplayconfig?.subplots?.length ?? 0) > 0
@@ -187,16 +175,10 @@ const displayTypeItems = computed<DisplayTypeItem[]>(() => {
   const noTooltip = props.informationContent === null
   const tooltipDefined = props.informationContent !== undefined
 
-  const chartEnabled = getDisplayEnabled(
-    props.settings.timeSeriesChart.enabled,
-    props.settings.timeSeriesChart.locationEnabledAttribute,
-  )
-  const tableEnabled = getDisplayEnabled(
-    props.settings.timeSeriesTable.enabled,
-    props.settings.timeSeriesTable.locationEnabledAttribute,
-  )
+  const chartEnabled = props.settings.timeSeriesChart.enabled
   const elevationChartEnabled =
     props.settings.verticalProfileChart.enabled && elevationChartsDefined
+  const tableEnabled = props.settings.timeSeriesTable.enabled
   const metaDataEnabled = props.settings.metaDataPanel.enabled && tooltipDefined
   return [
     {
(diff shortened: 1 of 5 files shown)


src/lib/topology/componentSettings/chartSettings.ts
```diff
@@ -1,7 +1,16 @@
 import type { DeepRequired } from '@/lib/utils/types'
 import type { ChartsSettings as PiChartsSettings } from '@deltares/fews-pi-requests'
 
-export type ChartsSettings = DeepRequired<PiChartsSettings>
+type PiChart = NonNullable<PiChartsSettings['timeSeriesChart']>
+type PiTimeSeriesTable = NonNullable<PiChartsSettings['timeSeriesTable']>
+
+export type ChartsSettings = Omit<
```
wkramer, line 7: Can you update to `"@deltares/fews-pi-requests": "^2.0.1"` in the `package.json`. That should include these type definitions.

src/components/timeseries/TimeSeriesWindowComponent.vue
```diff
@@ -94,6 +95,7 @@ interface Props {
   informationContent?: string | null
   filter?: filterActionsFilter | timeSeriesGridActionsFilter
   settings: ChartsSettings
+  location?: Location
```
wkramer, line 98: We should consider what is the correct behavior when multiple locations are selected. Currently only the attributes of the first selected location are applied.

src/components/timeseries/TimeSeriesWindowComponent.vue
```diff
@@ -166,6 +168,16 @@ watch(
   { immediate: true },
 )
 
+function getDisplayEnabled(
+  settingsEnabled: boolean,
+  attributeId: string | undefined,
+): boolean {
+  if (!attributeId) return settingsEnabled
+  const attr = props.location?.attributes?.find((a) => a.id === attributeId)
+  if (attr === undefined) return settingsEnabled
```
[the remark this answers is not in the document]
wkramer: I will check with @bogaardt what is the desired behavior. I think the current implementation is what we discussed.

wkramer (changes requested):

src/lib/topology/componentSettings/chartSettings.ts
TGehrmannHydrotec: Updating leads to several problems in various files. 
The usage of updated ChartsSettings asks for changes in a couple of files and ComponentSettings in even more files because of changed attribute types.
I checked the main and you're still using the old version there aswell if I'm correct. 
We're not sure if you're going to update yourself or you want us to do it.

src/components/timeseries/TimeSeriesWindowComponent.vue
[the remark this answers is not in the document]
TGehrmannHydrotec: I didn't know it was possible to select multiple locations. Do you want to discuss or will you come back to us when you decided about the desired behavior for multiple selected locations? Anton and I discussed it today but haven't managed to find a good solution yet.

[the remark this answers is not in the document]
wkramer: You can CTRL click locations to select multiple.
FEWS can also have parent/child locations. Clicking a parent will select all child locations.
In the Web OC the url/route will have comma separated list of `locationIds`.

src/lib/topology/componentSettings/chartSettings.ts
wkramer: I will fix this. I had a check an the nicest solution is to fix the issue in OpenAPI specification and regenerate the types in fews-pi-requests library. I will do this early next week.

wkramer:
We need another approach due to the multiple locations issue:

### Move the `getDisplayEnabled` function 

Move
```typescript
function getDisplayEnabled(
  settingsEnabled: boolean,
  attributeId: string | undefined,
)
```
out of the `TimeSeriesComponent.vue` to `SpatialDisplay.vue`.

### Restore `TimeSeriesComponent`

We will keep current version (i.e `main` branch version) of `TimeSeriesComponent.vue`. This already checks 
`settings.timeSeriesChart.enabled`, `settings.verticalProfileChart.enabled`, etc.


### Use current `settings` prop 

In `SpatialDisplay.vue` we pass the `settings` object to  `SpatialTimeSeriesDisplay` -> `TimeSeriesComponentWindow` -> `TimeSeriesComponent`.


### Update `settings.timeSeriesChart.enabled` based on selected locations

In `SpatialDisplay.vue`  if `settings.timeSeriesChart.locationEnabledAttribute` is set, determine for provided `locationIds` 

Update to the `getDisplayEnabled` function to loop over the locations.
```typescript
function getDisplayEnabledFromLocationAttributes(
  locations: Location[],
  attributeId: string | undefined,
  default: boolean,
)
```
If `default = false` and any location has an attribute
```
{ attributeId: true }
```
return `true`

If `default = true` and any location has an attribute
```
{ attributeId: false }
```
return `false`.

TGehrmannHydrotec:
I implemented the requested changes, they were pushed.

wkramer:
Implemented in https://github.com/Deltares/fews-web-oc/pull/2013

PR #1956, opened 2026-04-29
refactor: code quality issues (2)

wkramer:
### Description

Resolve code quality issues detected by SonarQube.

### DIFF
--- a/src/components/download/TimeSeriesFileDownloadComponent.vue
+++ b/src/components/download/TimeSeriesFileDownloadComponent.vue
@@ -6,7 +6,7 @@
       }}</v-card-title>
       <v-card-text>
         <v-text-field
-          v-model="fileName"
+          v-model="fileNameInput"
           :label="t('download.fileName')"
           variant="underlined"
           density="compact"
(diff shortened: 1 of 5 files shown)


src/components/download/TimeSeriesFileDownloadComponent.vue
```diff
@@ -163,13 +163,17 @@ watchEffect(() => {
...
+const fileNameInput = ref('timeseries')
+
+watch(
+  () => showDialog.value,
+  (newValue) => {
+    if (!newValue) return
+    const FILE_FORMAT_DATE_FMT = 'yyyyMMddHHmmss'
+    const defaultDateTimeString = DateTime.now().toFormat(FILE_FORMAT_DATE_FMT)
+    fileNameInput.value = `timeseries_${defaultDateTimeString}`
+  },
```
ceesvoesenek, line 175: Maybe extract this to a function so the intent is clear. Presumably it sets a default filename when the dialog is opened? Should the filename always reset when we close and open the dialog?

src/components/tasksdisplay/WhatIfDisplay.vue
```diff
@@ -403,6 +355,42 @@ async function submit() {
   }
 }
 
+async function createScenario(
```
ceesvoesenek: This name is a bit unclear I think? It's submitting a what-if scenario, not really creating it? Maybe `submitWhatIfScenario`?

src/components/tasksdisplay/WhatIfDisplay.vue
```diff
@@ -403,6 +355,42 @@ async function submit() {
...
+  }
+  const scenarioResult = await postWhatIfScenario(scenarioFilter)
+  if (scenarioResult.status === 'error') {
+    showErrorMessage(scenarioResult.error)
+    return null
+  }
+  return scenarioResult.data.id
+}
+
+function handleWorkflowSuccess(result: string | undefined): void {
```
ceesvoesenek: This name is also a bit unclear, the workflow itself is not successful, just submitting it? It is a bit difficult to find a good name for this function because it does three unrelated things. Maybe it should be cut up into three functions?

ceesvoesenek (commented):
Looks good, mainly some comments on extracting functions from the what-if-scenario-function that was probably flagged as too big.

src/components/download/TimeSeriesFileDownloadComponent.vue
wkramer: That is valid point but should be refactored in a separate issue (DWO-1314)

PR #1961, opened 2026-04-30
Await until user is ready in transformRequest

### LINKED ISSUE #1960: DWO-1313: Runtask menu empty, but with refresh shows data
https://issuetracker.deltares.nl/browse/DWO-1313

hvangeffen:
### Description
Fixes workflows being fetched before user being availabe for its token

### Checklist
    - `rel: new feature`
    - `rel: improvement`
    - `rel: fixes`
    - `rel: ignore`


src/services/authentication/AuthenticationManager.ts
```diff
@@ -6,63 +6,70 @@ import { mergeHeaders } from '@/lib/requests/transformRequest'
...
   userManager!: UserManager
   private user: User | null = null
-  private initPromise: Promise<void> | null = null
+  private userReady: Promise<void> | null = null
 
-  async init(settings: UserManagerSettings): Promise<void> {
-    if (!this.initPromise) {
-      this.initPromise = (async () => {
+  init(settings: UserManagerSettings): void {
+    this.userReady = new Promise(async (resolve, reject) => {
```
ceesvoesenek, line 12: It seems that this weird `Promise` property could just be a conditional section in `getUser`? That seems a bit simpler than having a `Promise` as a property in this class

src/services/authentication/AuthenticationManager.ts
```diff
@@ -6,63 +6,70 @@ import { mergeHeaders } from '@/lib/requests/transformRequest'
...
+    if (!user) return
+    if (!user.expired) return
+
+    try {
+      this.user = await this.userManager.signinSilent()
+    } catch (error) {
+      console.error('Silent sign-in failed:', error)
+      // Redirect to interactive login as a fallback
+      await this.userManager.removeUser()
+      const base = `${import.meta.env.BASE_URL}`
```
ceesvoesenek, line 48: No need for the template string

src/services/authentication/AuthenticationManager.ts
```diff
@@ -6,63 +6,70 @@ import { mergeHeaders } from '@/lib/requests/transformRequest'
...
+      const redirect = window.location.pathname.slice(base.length)
+      window.location.href = base + 'login?redirect=' + redirect
```
ceesvoesenek, line 50: Maybe a `URL` constructor?

ceesvoesenek (commented):
Looks good!

I think initialising the user could be simpler though

src/services/authentication/AuthenticationManager.ts
hvangeffen: I still want to have a single promise which all getUser requests await so i kept it the same

src/services/authentication/AuthenticationManager.ts
hvangeffen: Can't easily test this so i will keep it the same

ceesvoesenek (approved):

PR #1969, opened 2026-05-05
Improve data analysis display and add brush chart to it

### LINKED ISSUE #1968: DWO-1275: Brush chart on data analysis display

### LINKED ISSUE #1975: DWO-1290: Verbeteringen Data Analysis Display
https://issuetracker.deltares.nl/browse/DWO-1290

cees-wolfs-vt:
Fixes #1968 and #1975


src/components/analysis/AnalysisAddToButton.vue
```diff
@@ -3,8 +3,8 @@
     :disabled="numberOfNewCharts === 0"
     text="Create new charts"
     :loading="loadingNewCharts"
-    class="mr-2"
     @click="emit('addToChart')"
+    class="mr-2"
```
ceesvoesenek, line 6: Nitpick: isn't it a general convention to put event listener attributes last? Why was this moved?

src/components/analysis/AnalysisCollection.vue
```diff
@@ -10,6 +10,40 @@
     density="compact"
     max-width="250"
   >
+    <template #item="{ item, props }">
+      <v-list-item v-bind="props" :title="item.title">
+        <template #append>
+          <v-menu :close-on-content-click="false">
```
ceesvoesenek: This feels like a `v-dialog` rather than a `v-menu`?

src/components/analysis/AnalysisCollection.vue
```diff
@@ -72,4 +106,15 @@ function addCollection(): void {
   newCollectionName.value = ''
   dialog.value = false
 }
+
+function deleteCollection(collectionName: string) {
+  const index = props.collections.findIndex((c) => c.name === collectionName)
+  if (index !== -1) {
```
ceesvoesenek: Can be an early return to save some indentation

src/components/analysis/AnalysisCollection.vue
```diff
@@ -72,4 +106,15 @@ function addCollection(): void {
...
+    props.collections.splice(index, 1)
+    if (props.collections.length === 0) {
+      props.collections.push(createCollection('Default', props.config))
+    }
+    selectedCollectionName.value = props.collections[0].name
```
ceesvoesenek: Why do we always reset the selected collection upon deletion? That does not seem necessary in many cases?

src/components/analysis/AnalysisDataDownload.vue
```diff
@@ -0,0 +1,98 @@
+<template>
```
ceesvoesenek, line 1: In general, components with a single root element are a bit easier to work with. Attribute inheritance (e.g., `class`) is a bit weird with multiple root elements. Maybe just wrap it in a `<div>`?

src/components/analysis/AnalysisDataDownload.vue
```diff
@@ -0,0 +1,98 @@
...
+  <v-menu>
+    <template #activator="{ props }">
+      <v-btn v-bind="props" size="small" variant="tonal" :disabled>
```
ceesvoesenek: I had never realised this abbreviated attribute syntax existed!

src/components/analysis/AnalysisDataDownload.vue
```diff
@@ -0,0 +1,98 @@
...
+  const viewPeriod = {
+    startTime: convertJSDateToFewsPiParameter(props.startTime),
+    endTime: convertJSDateToFewsPiParameter(props.endTime),
+  }
+  const headers = await authenticationManager.getAuthorizationHeaders()
+
+  const fileType = selectedFileType.value
+
+  const date = new Date()
+  const dateString = date.toISOString().replace(/[-:T]/g, '').slice(0, 14)
```
ceesvoesenek: 😢 so sad that we need to do this.

Is there not a utility function for this somewhere? If not, there probably should be, we're doing this kind of thing all over the place (because FEWS can't accept regular ISO8601...)

src/components/analysis/AnalysisDataDownload.vue
```diff
@@ -0,0 +1,98 @@
...
+async function downloadFileSafe(
+  url: string,
+  fileName: string,
+  documentFormat: DocumentFormat,
+  headers: Headers,
+) {
+  try {
+    await downloadFileAttachment(url, fileName, documentFormat, headers)
+  } catch (error) {
+    if (error instanceof Error) {
```
ceesvoesenek: Do we not always want to show an error when something is thrown? Maybe just `const message = error.message ?? 'Unknown error occurred'`?

src/components/analysis/AnalysisDateRange.vue
```diff
@@ -0,0 +1,92 @@
...
+  <v-menu :close-on-content-click="false">
+    <template #activator="{ props }">
+      <v-btn v-bind="props" :text="getDateRangeString()" />
+    </template>
+
+    <v-card flat width="400">
+      <v-card-text>
+        <div class="d-flex ga-2">
+          <v-date-input
+            v-model="collection.settings.startTime"
```
ceesvoesenek: I'm not a huge fan of components modifying their properties. Can we not define a few `v-model` variables here instead?

src/lib/analysis/filter.ts
```diff
@@ -28,6 +29,16 @@ export async function createNewChartsForFilters(
   return charts.filter((chart) => chart !== undefined)
 }
 
+async function fetchActionAndFullAction(filter: filterActionsFilter) {
+  return {
+    action: await fetchActions(baseUrl, filter),
+    fullAction: await fetchActions(baseUrl, {
```
ceesvoesenek: Can be done concurrently

src/lib/download/download.ts
```diff
@@ -0,0 +1,66 @@
...
+  isDataDownloadFilter,
+  isFilterActionsFilter,
+  isTimeSeriesFilter,
+  isTimeSeriesGridActionsFilter,
+  isTimeSeriesTopologyActionsFilter,
+} from '@/lib/filters'
+
+export function getDownloadFileUrl(
+  baseUrl: string,
+  filter: Filter,
```
ceesvoesenek, line 15: I think the API here would be a bit nicer if you restrict the types to only supported types instead of all filters here, instead of relying only on a runtime error.

src/lib/filters/index.ts
ceesvoesenek, line 1: Out of scope, but shouldn't these functions be in `@deltares/fews-pi-requests`?

src/components/analysis/AnalysisDateRange.vue
```diff
@@ -0,0 +1,92 @@
...
+            :disabled="collection.settings.liveUpdate.enabled"
+          />
+        </div>
+        <v-switch
+          v-model="collection.settings.liveUpdate.enabled"
+          label="Live Update"
+          hide-details
+        />
+        <div class="d-flex ga-2">
+          <v-number-input
```
ceesvoesenek, line 42: These fields can be left empty? What's the behaviour then?

ceesvoesenek (commented):
Some relatively minor comments to polish this a bit. Looks good overall!

src/components/analysis/AnalysisDataDownload.vue
hvangeffen: If we need it later we could always add a div. Currently I feel like we don't need to change the dom for this.

src/components/analysis/AnalysisDataDownload.vue
hvangeffen: Moved this and the other one to a util function

src/components/analysis/AnalysisDataDownload.vue
hvangeffen: Makes a long list of these quite a bit more readable

src/components/analysis/AnalysisDateRange.vue
hvangeffen: At some layer in the code we will have to point to the collection state.
The sole purpose of this component is to compose together more general components (like vuetify inputs) and manage a specific piece of analysis collection state.

I personally feel like changing it to something like this makes it less readable:

```ts
const startTime = defineModel<Date>('startTime', { required: true })
const endTime = defineModel<Date>('endTime', { required: true })
const liveUpdateEnabled = defineModel<boolean>('liveUpdateEnabled', {
  required: true,
})
const liveUpdateDaysBeforeNow = defineModel<number>('liveUpdateDaysBeforeNow', {
  required: true,
})
const liveUpdateDaysAfterNow = defineModel<number>('liveUpdateDaysAfterNow', {
  required: true,
})
```
with in the parent
```vue
          v-model:startTime="selectedCollection.settings.startTime"
          v-model:endTime="selectedCollection.settings.endTime"
          v-model:liveUpdateEnabled="selectedCollection.settings.liveUpdate.enabled"
          v-model:liveUpdateDaysBeforeNow="selectedCollection.settings.liveUpdate.daysBeforeNow"
          v-model:liveUpdateDaysAfterNow="selectedCollection.settings.liveUpdate.daysAfterNow"
```

src/lib/download/download.ts
hvangeffen: It already exhausts filter: 
image
But according to typescript if I omit it:
image
I ended up making `timSeriesUrl` the default return, so if in the future more are added we get an error and have to add it here.

src/lib/filters/index.ts
hvangeffen: `@deltares/fews-pi-requests` does not currently export any type guard functions

src/components/analysis/AnalysisDateRange.vue
hvangeffen: Empty means 0 days or now (new Date())

src/components/analysis/AnalysisDateRange.vue
ceesvoesenek: Hmm yes that's pretty noisy. Can't we `v-model` the collection settings instead as a middle ground?

src/components/analysis/AnalysisDataDownload.vue
ceesvoesenek: It's pretty surprising behaviour if you're trying to add a class to this component without knowing its internals though. A wrapping `<div>` is basically free in terms of performance.

src/components/analysis/AnalysisDateRange.vue
ceesvoesenek: It's replacing negative values with 0, but not the empty string. Explicitly replacing with zero makes the behaviour a bit more obvious I think.

ceesvoesenek:
Two more small UI points, now that we're doing big improvements:

It's really not obvious you can click this date range:
image

We should probably make it a bit more buttony (maybe add a chevron?) so users are more likely to find out they can click it.

I like it a bit more if these switches turn into the primary colour if they are enabled, it makes it a bit more obvious what state you are currently in.

image

src/lib/filters/index.ts
ceesvoesenek: It doesn't, but I feel it should!

These functions seem like general purpose library function, and they are heavily tied to the data structures in `@deltares/fews-pi-requests`. If one of the properties used as a discriminator gets added to another filter, these functions are no longer correct.

src/components/analysis/AnalysisDataDownload.vue
hvangeffen: Added a div

src/components/analysis/AnalysisDateRange.vue
hvangeffen: Made it v-model:settings

src/lib/filters/index.ts
hvangeffen: Would be nice in the future yes, might be a good consideration for the coming pi request changes

src/components/analysis/AnalysisDateRange.vue
hvangeffen: Made null go to 0 here

PR #2001, opened 2026-05-28
Fix task overview only showing tasks for currently selected node

hvangeffen:
### Checklist
    - `rel: new feature`
    - `rel: improvement`
    - `rel: fixes`
    - `rel: ignore`
- [x] Update documentation.
- [x] Update tests.

### DIFF
--- a/src/components/sidepanel/TaskOverviewSidePanel.vue
+++ b/src/components/sidepanel/TaskOverviewSidePanel.vue
@@ -144,7 +144,6 @@ const period = ref<RelativePeriod | null>({
   endOffsetSeconds: 0,
 })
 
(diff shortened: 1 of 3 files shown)


src/components/sidepanel/TaskOverviewSidePanel.vue
```diff
@@ -160,7 +160,7 @@ const {
   selectedWorkflowIds,
   selectedTaskStatuses,
   selectedUserId,
-  () => props.topologyNode?.id,
+  undefined,
   true,
```
ceesvoesenek: It seems that this parameter is never `false`... Maybe we can remove it so we can get rid of the nasty explicit `undefined`?

ceesvoesenek (approved):

PR #2010, opened 2026-06-02
 Fix vulnerabilites in sonarcube

### LINKED ISSUE #2002: DWO-1310: Fix vulnerabilites in SonarCube

wkramer:
### Description

Only use full path to Git executable, to prevent malicious replacement of the executables on PATH.

### DIFF
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -2,11 +2,33 @@ import { defineConfig, loadEnv } from 'vite'
 import vue from '@vitejs/plugin-vue'
 import vuetify from 'vite-plugin-vuetify'
 import { dirname, resolve } from 'node:path'
-import { execSync } from 'node:child_process'
+import { execFileSync } from 'node:child_process'
+import { existsSync } from 'node:fs'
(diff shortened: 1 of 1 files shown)


vite.config.ts
```diff
@@ -1,12 +1,34 @@
...
-import { dirname, resolve } from 'path'
-import { execSync } from 'child_process'
-import { fileURLToPath } from 'url'
+import { dirname, resolve } from 'node:path'
+import { execFileSync } from 'node:child_process'
+import { existsSync } from 'node:fs'
+import { fileURLToPath } from 'node:url'
 
-const commitHash = execSync('git rev-parse --short HEAD').toString()
-const commitTag = execSync('git tag --points-at HEAD').toString()
```
hvangeffen, line 9: Would something like this also fix sonarcube? `execFileSync('git', ['rev-parse', '--short', 'HEAD'])`

wkramer: No, `git` is not an existing file.

PR #2015, opened 2026-06-03
Implement plot weight for time series chart and update chart series configuration

### LINKED ISSUE #2005: DWO-1304: Take plot weight for time series window into account

ShomailaF:
image


src/components/timeseries/TimeSeriesComponent.vue
```diff
@@ -311,6 +301,57 @@ const subplots = computed(() =>
   ),
 )
 
+const subplotLayoutWeights = computed(() => {
```
wkramer: The plotWeight should be used for the axis height. So it should compensate for the axis margins. The brush chart height should not be included in the calculation for the plotWeight.

wkramer (approved):

PR #2110, opened 2026-09-01
Add permissions toggle control

ceesvoesenek:
### Description
The FEWS PI REST Web Services allows specifying the `fews-ws-permissions-excludes` header to disabling specific permissions that are normally assigned to the user, to simulate what the interface looks like for less privileged users.

This PR adds a control to the about page to set these excluded permissions.

### Checklist
- [ ] Update documentation.


src/components/permissions/ExcludedPermissionsControl.vue
```diff
@@ -0,0 +1,84 @@
...
+  <v-list-group v-if="hasAssignedPermissions" subgroup>
+    <template #activator="{ props }">
+      <v-list-item v-bind="props" :title="t('userSettings.permissions')" />
+    </template>
+
+    <v-list-item
+      v-for="(permissionId, index) in permissionsStore.assignedPermissionIds"
+      :key="index"
+    >
+      {{ permissionId }}
```
hvangeffen, line 11: I saw that the backend now also sometimes returns a name for a permission:
```typescript
export interface Permission {
  /**
   * id of the permission
   */
  id: string;
  /**
   * Optional name of the permission
   */
  name?: string;
  /**
   * True if the permission is assigned to the current user.
   */
  assigned: boolean;
}
```
Will require an update of pi-requests

ceesvoesenek: Updated `fews-pi-requests` and using this name if it is available, with a fallback to the ID.

src/components/permissions/ExcludedPermissionsControl.vue
```diff
@@ -0,0 +1,106 @@
...
+        <v-list-item
+          v-bind="props"
+          :title="t('userSettings.permissions')"
+          :subtitle="subtitle"
+        />
+      </template>
+
+      <v-list-item
+        v-for="(permissionId, index) in permissionsStore.assignedPermissionIds"
+        :key="index"
```
hvangeffen: Would using permissionId as the key here be wrong?

src/views/auth/LoginComponent.vue
```diff
@@ -29,18 +35,20 @@ import {
...
 
-import { useI18n } from 'vue-i18n'
-
 const { t } = useI18n()
-
 const route = useRoute()
-const initials = ref('')
+
+const PLACEHOLDER_NAME = 'User'
+const hasUserManager = !!authenticationManager?.userManager
```
hvangeffen: Might be nice to move this to authenticationManager

src/views/auth/LoginComponent.vue
```diff
@@ -51,19 +59,23 @@ function setUser() {
     })
 }
 
-authenticationManager.userManager.events.addUserLoaded(() => {
-  requiresLogin.value = false
-  setUser()
-})
+if (hasUserManager) {
+  authenticationManager.userManager.events.addUserLoaded(() => {
```
hvangeffen: Stuff like this in this comopnent also seems like we could abstract away into the authenticationManager

src/components/permissions/ExcludedPermissionsControl.vue
```diff
@@ -0,0 +1,106 @@
...
+            color="primary"
+            size="small"
+            variant="flat"
+            :text="t('common.apply')"
+            :disabled="disableExcludePermissions || !hasPendingChanges"
+            @click="applyPendingChanges()"
+          />
+        </div>
+      </v-list-item>
+    </v-list-group>
```
hvangeffen: This still seems to be expandable and viewable when `Simulating disabled permissions is not allowed by the server.` and there are no permissions returned: 

Image

hvangeffen: Do we even want to show the permissions UI at all when it is disabled?

src/stores/permissions.ts
```diff
@@ -0,0 +1,166 @@
...
+      try {
+        const response = await piProvider.getPermissions()
+        permissions.value = response.permissions ?? []
+      } finally {
+        hasLoaded.value = true
+      }
+    }
+
+    async function checkExcludedPermissionsHeaderAllowed(): Promise<void> {
+      const baseUrl = configManager.get('VITE_FEWS_WEBSERVICES_URL')
```
hvangeffen: Might be nice to do this once at the store level

src/stores/permissions.ts
```diff
@@ -0,0 +1,166 @@
...
+      const headers = await getRequestHeaders(true)
```
hvangeffen: Would be nice to have `disablePermissionExcludes` in an options object so this automatically becomes more readable.

src/stores/permissions.ts
```diff
@@ -0,0 +1,166 @@
...
+      headers.set(FEWS_PERMISSION_EXCLUDES_HEADER, 'dummy')
+
+      // Try an options request with the permission excludes header on the
+      // version endpoint to see whether the server accepts the header. There
+      // might be several reasons why it might not, e.g. web server
+      // configuration.
+      const versionUrl = `${baseUrl}/version`
```
hvangeffen: To prevent url problems later it would be nice to still construct a `piProvider` here and use `versionUrl` here.

src/views/auth/LoginComponent.vue
```diff
@@ -2,15 +2,20 @@
...
             {{ initials }}
           </v-btn>
         </template>
         <v-list density="compact">
-          <v-list-item>{{ name }}</v-list-item>
-          <v-list-item @click="logout">
+          <v-list-item>
+            <div class="d-flex align-center">
+              <span class="user-avatar mr-2">{{ initials }}</span>
+              <span class="user-name">{{ name }}</span>
```
hvangeffen: Might be nice to use a `v-avatar` for the initials in both the button and the span with class here.
This can be done as such:
```vue
    <v-avatar color="red">
      <span class="text-headline-small">CJ</span>
    </v-avatar>
```

See: https://vuetifyjs.com/en/components/avatars/#default

src/App.vue
```diff
@@ -1,7 +1,11 @@
 <template>
   <Suspense>
     <template #default>
-      <component style="height: 100%" :is="layoutComponent"></component>
+      <component
+        style="height: 100%"
+        :is="layoutComponent"
+        :key="permissionsStore.excludedPermissionsKey"
```
hvangeffen, line 7: Do we still need this since the user changes their permissions in the `/about` page?

src/stores/permissions.ts
```diff
@@ -0,0 +1,166 @@
...
+
+    const assignedPermissionIds = computed<string[]>(() =>
+      permissions.value
+        .filter((permission) => permission.assigned)
+        .map((permission) => permission.id),
+    )
+    const excludedPermissionIds = ref<string[]>([])
+
+    const excludedPermissionsKey = computed<string>(() => {
+      if (excludedPermissionIds.value.length === 0) return 'none-excluded'
```
hvangeffen, line 40: Would be nice as a const on top

src/stores/permissions.ts
```diff
@@ -0,0 +1,166 @@
...
+            // Otherwise, add it to the list of excluded permissions.
+            newExcludedPermissionIds.push(currentPermissionId)
+          }
+        },
+      )
+      // Update the entire array of excluded permissions at once; this will
+      // result in a different permissions key which can be used to trigger a
+      // reload of (parts of) the application.
+      excludedPermissionIds.value = newExcludedPermissionIds
+    }
```
hvangeffen, line 88: Would a set work here as well?: 

```ts
function setPermissions(updatedPermissions: Record<string, boolean>): void {
  const excluded = new Set(excludedPermissionIds.value)

  Object.entries(updatedPermissions).forEach(
    ([permissionId, isActive]) => {
      if (isActive) {
        excluded.delete(permissionId)
      } else {
        excluded.add(permissionId)
      }
    },
  )

  excludedPermissionIds.value = [...excluded]
}
```

src/stores/permissions.ts
hvangeffen: Same for `createTransformRequestFn`

src/components/permissions/ExcludedPermissionsControl.vue
ceesvoesenek: No that would be fine, good idea!

src/views/auth/LoginComponent.vue
ceesvoesenek: I've kept these things the same as the previous state of the PR, which was already reviewed, I think?

I'll see if I can test this properly, and if I can move it there

src/components/permissions/ExcludedPermissionsControl.vue
ceesvoesenek: Viewing permissions might still be of interest to users?

I agree it should not be visible if there are no assigned permissions though!

src/stores/permissions.ts
ceesvoesenek: Good point

src/components/permissions/ExcludedPermissionsControl.vue
```diff
@@ -0,0 +1,106 @@
...
+            :disabled="disableExcludePermissions || !hasExcludedPermissions"
+            @click="resetPermissions()"
+          />
+          <v-btn
+            color="primary"
+            size="small"
+            variant="flat"
+            :text="t('common.apply')"
+            :disabled="disableExcludePermissions || !hasPendingChanges"
+            @click="applyPendingChanges()"
```
hvangeffen, line 40: On apply the appBarColor seems to dissapear. 
This happens because on remount this watch does not trigger:
```ts
watch(
  () => theme.global.current.value,
  () => {
    nextTick(updateAppBarStyles)
  },
)
```
Since the theme is not changing.
This watch seems in general quite dependant on theme loading slower than itself.

src/stores/permissions.ts
ceesvoesenek: Yes, much better!

src/components/permissions/ExcludedPermissionsControl.vue
ceesvoesenek: Ohh exciting! Good catch

src/stores/permissions.ts
ceesvoesenek: Good idea, I don't like passing `undefined` and the impact on the rest of the code base is minimal

src/views/auth/LoginComponent.vue
ceesvoesenek: I've just realised that all these changes to the login component are not necessary now that we've moved the permissions to the about page.

I'll revert them.

hvangeffen (approved):

src/App.vue
ceesvoesenek: The content of the layout component may change based on the permissions, so a remount is probably the safest option

